PR #175, opened 2013-11-12
Proposed INVITE w/o SDP UAC behavior

wakamoleguy:
We've been working on this as a custom feature, so I thought I'd push it back to the community for input.  We've implemented INVITE w/o SDP as per RFC3261.

Implementation notes:

This is mostly built on the existing RTCSession structure.  The media stream is still requested before inviting, and the stream is carried around on the session until it is needed.  The application probably wants to add an `OfferToReceiveVideo: false` constraint when making audio-only calls.

RTCMediaHandler was modified to store RTCConstraints beyond initial creation, as we found browsers mostly didn't respect the setting until the createAnswer() call, and the session didn't want to carry it around for that long.

Thoughts?

### DIFF
--- a/src/RTCSession.js
+++ b/src/RTCSession.js
@@ -44,6 +44,7 @@ RTCSession = function(ua) {
   this.dialog = null;
   this.earlyDialogs = {};
   this.rtcMediaHandler = null;
+  this.mediaStream = null;
 
   // Session Timers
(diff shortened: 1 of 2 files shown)


jmillan:
Closing. PR does not apply anymore :-)

PR #236, opened 2014-07-16
Screen sharing for outgoing calls

mEkblom:
Call constraints parameter accept attribute screen: true/false for screen sharing. Screen sharing only works in Chrome using https with Chrome flag screen-capture set to true.

### DIFF
--- a/src/RTCSession.js
+++ b/src/RTCSession.js
@@ -1326,8 +1326,46 @@ RTCSession.prototype.sendInitialRequest = function(mediaConstraints, RTCOfferCon
  this.receiveResponse = this.receiveInviteResponse;
 
  if (mediaStream) {
+    // Use already exiting media stream
    userMediaSucceeded(mediaStream);
+
+  } else if(mediaConstraints.screen) {
+    // Try to share screen (will require Chrome with screen-capture flag and https)
+    this.rtcMediaHandler.getUserMedia(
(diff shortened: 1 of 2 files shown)


shaksK:
Martin,

I am trying to patch your code with an older version of JSIP and to try and get screen share working on Chrome. I did install the screen share plugin. However I can't get it to work. Can you please suggest what sort of modifications might be needed to your initial pull request ?
thank you,
Shaks.

mEkblom:
Hi shaksK,

sadly native support for screen sharing has been removed from recent versions of Chrome. This means that this patch will not work anymore. However, the internal logic for screen capture is still in Chrome but you will need to use (or make your own) Chrome [plugin] (https://chrome.google.com/webstore/detail/webrtc-desktop-sharing/nkemblooioekjnpfekmjhpgkackcajhg) to reach this functionality.

I don't have the time right now to provide a new patch but I can confirm it works fine (basically the screen id needs to be added to the constraints and wrapped in a callback for the plugin). Also the JsSIP project goes in another direction now (breaking out media capabilities) so it does not make much sense to include it.

shaksK:
Martin,
Thank you very much for your input. It is indeed sad that chrome has
discontinued native screen sharing. Anyway I will perhaps migrate to a
different stack.
Sharath

PR #255, opened 2014-10-17
connectionFailed event and connection_autorecovery flag

rusekr:
Event `connectionFailed` to separate connection failure from normal disconnection process and possibility to disable auto connection recovery with `connection_autorecovery` flag. With this any wrapper to JsSIP can control connection recovery itself syncing it with other modules.

If you prefer not to have `connectionFailed` event I can separate this feature..

### DIFF
--- a/src/Transport.js
+++ b/src/Transport.js
@@ -250,7 +250,7 @@ Transport.prototype = {
 
     this.reconnection_attempts += 1;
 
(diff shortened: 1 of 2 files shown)


c35xr:
why this was closed? exists a way to define a  custom the Auto-Recovery system?

PR #264, opened 2014-10-31
Filter for loops

sanyatuning:
Hi

I having trouble while using JsSIP.
There are many [unfiltered "for in" loops (for arrays)](http://www.jshint.com/docs/options/#forin) what breaks my code.
Can you fix this?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -117,6 +117,7 @@ module.exports = function(grunt) {
 				// DOC: http://www.jshint.com/docs/options/
 				curly: true,
 				eqeqeq: true,
+				forin: true,
 				immed: true,
 				latedef: true,
 				newcap: true,


jmillan:
Closing for inactivity.

sanyatuning:
These
https://travis-ci.org/versatica/JsSIP/builds/39574860

I turned on the check in your jshint config

ibc:
@sanyatuning you have posted a link to Travis, not sure what you mean with that.

sanyatuning:
there are many errors in jshint task output like this
```
src/SIPMessage.js
    182 |    for (header in this.headers) {
             ^ The body of a for in should be wrapped in an if statement to filter unwanted properties from the prototype.
   ```

But nowdays I no longer use JsSIP so you may ignore this request

jmillan:
Hi @sanyatuning,

We'll keep it open. Definitely we didn't manage the PR properly.

Thanks a lot.

ibc:
@sanyatuning Regardless we want to avoid JsSIP failing due to an app mangling the Array prototype, I don't think that you should apply jshint on external libraries but just in your own JS code. Otherwise, may your jshint also complain because we use spaces vs TAB?

PR #304, opened 2015-02-05
FIX: no events on incomming call

kvishnivetsky:


kvishnivetsky:
Thank's @jmillan , I'll fix this.

jmillan:
Let me close this pull request since it does not solve an issue. There is a documented way to set the event handlers for an incoming call.

Thanks a lot though @kvishnivetsky

ibc:
Agreed. Setting events if fully feasible for both incoming and outgoing calls. There is no issue at all, no limitation at all. Extending the API to allow more ways of doing the same is not desired.

PR #334, opened 2015-09-13
Don't ask for media in answers if audio and video constraints are not set

Rob--W:
0.6.8 added a similar feature, but only for outgoing calls.


ibc:
Makes sense, thanks a lot. I've added it in d20f227df46d2e099ecd215ab6d9f59089c9ae4e following the same format as in the outgoing calls.

PR #342, opened 2015-10-14
Incorrect logic in uri scheme parsing

adombrovsky:
If sip client uses sips protocol error:  
`error parsing "From" header field with value occurs`

Seems that parsing of uri scheme has incorrect logic. First of all must be parsed *sips* protocol and then *sip*

Example of header:

`
<sips:5000000246@host.com>;tag=Ka7al2s~f
`

### DIFF
--- a/lib/Grammar.js
+++ b/lib/Grammar.js
@@ -3059,18 +3059,7 @@ module.exports = (function(){
         var result0;
         var pos0;
         
-        if (input.substr(pos, 3).toLowerCase() === "sip") {
-          result0 = input.substr(pos, 3);
-          pos += 3;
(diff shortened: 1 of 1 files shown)


ibc:
Grammar.js is autogenerated. Changes must be done in Grammar.pegjs.

Anyhow, let @jmillan comment about "sips" usage in JsSIP as I think there were some issues...

jmillan:
Hi,

Thanks for reporting and sorry about the delay.

Please, do never edit Grammar.js since it's auto-generated from Grammar.pegjs file.

The issue is fixed in 0.7.14 tag.

PR #343, opened 2015-10-21
Correct contact headers for REGISTER

mr-tron:
Some backends (like huaweiATS9900) need Contact header with real login.


ibc:
That is a bug in that server. RFC 3261 clearly states that the SIP URI at the Contact may have any username chosen by the client, and we want that in JsSIP.

Please tell Huawei to fix their SIP server.

PR #344, opened 2015-10-21
hack_username_in_contact parameter

rusekr:
Logical addition to hack_ip_in_contact parameter. For some strict servers.
If we can use any string in contact why not to give user chance to set its own (:

Input values:
false - random string (default)
true - uri.user
string - string given by client config

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -834,6 +834,7 @@ UA.prototype.loadConfig = function(configuration) {
     * Value to be set in Via sent_by and host part of Contact FQDN
     */
     via_host: Utils.createRandomToken(12) + '.invalid',
+    uri_user: Utils.createRandomToken(8),
 
     // Password
     password: null,
@@ -860,6 +861,7 @@ UA.prototype.loadConfig = function(configuration) {
     hack_via_tcp: false,
     hack_via_ws: false,
     hack_ip_in_contact: false,
+    hack_contact_username: false,
 
     // Options for Node.
     node_websocket_options: {}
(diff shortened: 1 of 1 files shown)


ibc:
Let me review the PR and manage it on the next weeks. Thanks.

jmillan:
In fact setting the contact URI user value to the SIP URI user one would avoid the new configuration parameter and satisfy this requirement. Also, it looks natural that both fields have the same value.

```
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -992,7 +992,7 @@ UA.prototype.loadConfig = function(configuration) {
   this.contact = {
     pub_gruu: null,
     temp_gruu: null,
-    uri: new URI('sip', Utils.createRandomToken(8), settings.via_host, null, {transport: 'ws'}),
+    uri: new URI('sip', settings.uri.user, settings.via_host, null, {transport: 'ws'}),
     toString: function(options) {
       options = options || {};
```

What about that?

rusekr:
Hello!
About "Why not a simple contact_username.." - agree but need time to "patch" this patch. 

"hack" prefix was added to comply with "hack_ip_in_contact" because it needed for some hardware SIP proxes (huawei) and not needed by RFC.

About "settings.uri.user" - It was pull request from my collegue which suggested that: https://github.com/versatica/JsSIP/pull/343
But we now have two setups with third side SIP hardware proxies (we can`t cofigure it) one of theese needs normal phone number for authorization (<number>) and string "<number><some random code>" in contact uri for parallel registrations and other need the same except don`t need "<some random code>" in contact uri - therefore we need some distinct parameter to configure uri in contact.

jmillan:
>But we now have two setups with third side SIP hardware proxies (we can't cofigure it) one of theese needs normal phone number for authorization (number) and string "(number)(some random code)" in contact uri for parallel registrations and other need the same except don't need "(some random code)" in contact uri - therefore we need some distinct parameter to configure uri in contact.

That is a hack..

Then prepare a PR according to my note on `hack_contact_username` (yes, with the `hack_`) prefix.

Also indicate the SIP server manufacturer/model for which the hack is needed so it can be properly referenced in the doc.

rusekr:
Changed name for this option to `hack_contact_username`. 

About `SIP server manufacturer/model` - it's Huawei IMS Core (http://huawei.com/ru/products/core-network/singlecore/ims-core/index.htm , only russian page) configured by MGTS company (http://mgts.ru). And it's not hardware specific - it's configuration specific. But configuration which we can not change - our client, their SIP proxy.

jmillan:
No more hack parameters are needed thanks to the new 'contact_uri' UA configuration parameter. Just provide the Contact URI string you'd like JsSIP to use..

For example:
- numerical only user
- alphanumerical user
- ip type host
- name type in host
- 'ws' transport parameter even using WSS transport.
.., etc, etc

ibc:
> 'ws' transport parameter even using WSS transport. .., etc, etc

That's right. You probably meant "using the undocumented and non-standard `;transport=wss` when using secure WebSocket (hello Asterisk).

jmillan:
yes, definitely

PR #348, opened 2015-12-09
Enable echo cancellation and noise suppression

andrey-utkin:
I don't have the reference to source of this solution at hand, but to my knowledge this fixes echo issues in both Chrome and Firefox.


saghul:
I doubt any goog* prefixed optins have any meaning on Firefox. At any rate, these changes need to go in the real source, not in the generated one in dist/.

andrey-utkin:
I wonder about goog* too, but AFAIK Firefox uses some parts of webrtc implementation with Chrome, no?

saghul:
Yep, they do use the same WebRTC engine but AFAIK they have their own PeerConnection. Some links to documentation / code would help.

andrey-utkin:
The new commit makes it in lib/
Untested, will look at it later (or maybe you guys find time earlier).

ibc:
@andrey-utkin thanks a lot for pointing this trick out, but I'm against hardcoding vendor private stuff in JsSIP, even more when the user can already pass those constraints via the `mediaConstraints` object in both `ua.call()` and `session.answer()` methods. If we added those constraints to JsSIP then we'd have to track them forever, and I do not want such a responsibility.

Thanks a lot.

PR #351, opened 2015-12-29
Add 'replying' event so that applications can mangle SDP before response is sent.

pg94au:
This change would introduce a new 'replying' event, similar to the existing 'sending' event.  The 'replying' event allows the application to modify the body of the OK response to an INVITE in a manner similar to how the 'sending' event allows the body of the INVITE message to be modified.

Using this is similar to the 'sending' event.  ie:

session.on('replying', sessionReplyingHandler);

function sessionReplyingHandler(e) {
	e.response.body = // update with modifications as desired
}


ibc:
@jmillan may you please take a look to this PR? I don't see any issue on it but would like to know your thoughts.

pg94au:
Hi, sorry, can you hold off on this for just a bit?  I've just recently been looking at this issue again myself, and it looks like this and the existing sending event aren't enough to allow interception of the SDP when going on and off hold (ie. bitrate, codec selection, etc. revert back to unmodified defaults in this case).  I will post something soon in the group, when I get a chance.

ibc:
Indeed, the 'sending' event is just fired for the initial outgoing INVITE.

ibc:
The purpose of this PR is accomplished in #372 (once merged).

PR #359, opened 2016-02-11
Force wss protocol (hack_via_wss)

orszaczky:
JSSip has connection issues over wss protocol, so I added hack_via_wss that fixed our problem.

### DIFF
--- a/lib/Transactions.js
+++ b/lib/Transactions.js
@@ -62,6 +62,9 @@ function NonInviteClientTransaction(request_sender, request, transport) {
   else if (request_sender.ua.configuration.hack_via_ws) {
     via_transport = 'WS';
   }
+  else if (request_sender.ua.configuration.hack_via_wss) {
+    via_transport = 'WSS';
+  }
   else {
     via_transport = transport.server.scheme;
   }
@@ -166,6 +169,9 @@ function InviteClientTransaction(request_sender, request, transport) {
   }
   else if (request_sender.ua.configuration.hack_via_ws) {
     via_transport = 'WS';
+  }
+  else if (request_sender.ua.configuration.hack_via_wss) {
+    via_transport = 'WSS';
   }
   else {
     via_transport = transport.server.scheme;
@@ -358,6 +364,9 @@ function AckClientTransaction(request_sender, request, transport) {
   }
   else if (request_sender.ua.configuration.hack_via_ws) {
     via_transport = 'WS';
+  }
+  else if (request_sender.ua.configuration.hack_via_wss) {
+    via_transport = 'WSS';
   }
   else {
     via_transport = transport.server.scheme;
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -894,6 +894,7 @@ UA.prototype.loadConfig = function(configuration) {
     // Hacks
     hack_via_tcp: false,
     hack_via_ws: false,
+    hack_via_wss: false,
     hack_ip_in_contact: false,
 
     // Options for Node.
@@ -992,7 +993,7 @@ UA.prototype.loadConfig = function(configuration) {
   this.contact = {
     pub_gruu: null,
     temp_gruu: null,
-    uri: new URI('sip', Utils.createRandomToken(8), settings.via_host, null, {transport: 'ws'}),
+    uri: new URI('sip', Utils.createRandomToken(8), settings.via_host, null, {transport: (settings.hack_via_wss ? 'wss' : 'ws')}),
     toString: function(options) {
       options = options || {};
 
@@ -1073,6 +1074,7 @@ UA.configuration_skeleton = (function() {
       'display_name',
       'hack_via_tcp', // false
       'hack_via_ws', // false
+      'hack_via_wss', // false
       'hack_ip_in_contact', //false
       'instance_id',
       'no_answer_timeout', // 30 seconds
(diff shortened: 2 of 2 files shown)


ibc:
> JSSip has connection issues over wss protocol

Which "connection issues"? The fact that it fails with some non specified server does not mean that this is a bug in JsSIP. Please clarify.

ibc:
Also, why do you need this?:

```js
else if (request_sender.ua.configuration.hack_via_wss) {
    via_transport = 'WSS';
}
```

If the WebSocket server has `wss://` schema then JsSIP **already** sets `'SIP/2.0/WSS`:

```js
  if (request_sender.ua.configuration.hack_via_tcp) {
    via_transport = 'TCP';
  }
  else if (request_sender.ua.configuration.hack_via_ws) {
    via_transport = 'WS';
  }
  else {
    via_transport = transport.server.scheme;
  }
```

ibc:
In the other side, regarding the SIP URI `transport` parameter, the PR is just incorrect:

```js
uri: new URI('sip', Utils.createRandomToken(8), settings.via_host, null, {transport: (settings.hack_via_wss ? 'wss' : 'ws')}),
```

There is no `;transport=wss`. [RFC 7118](http://tools.ietf.org/html/rfc7118) (SIP over WebSocket) does not define it at all, in the same way that RFC 3261 does not define `;transport=tls` when the transport is TLS.

ibc:
@orszaczky any comment please?

orszaczky:
[Starting with Chrome 47](https://developers.google.com/web/updates/2015/10/chrome-47-webrtc?hl=en), getUserMedia() requests are only allowed from secure origins: HTTPS or localhost: jssip.js:23089 getUserMedia() no longer works on insecure origins. To use this feature, you ...
Chrome also won't allow mixed content, if the page is loaded on https:// it silently blocks ws:// and only allows wss:// connections.

When JSSIP is connected to an Asterisk server through https and wss, calls to the extension will fail (asterisk reports the extension unavailable) while it's registered, `sip show peers` shows status as unreachable:
1001/1001                 1.0.x.x                               D  Yes        Yes            1903     UNREACHABLE

Maybe it's not defined in the standards (and can be a Chrome only requirement), setting SIP URI transport parameter to 'wss' does solve the problem and the extension status becomes ok, and can be called without any further issues:
1001/1001                 1.0.x.x                               D  Yes        Yes            1990     OK (395 ms)

I didn't test this behavior in other browsers, it's not relevant for our current Chrome only project.

I can't confirm that the changes in Transactions.js are necessary, i just followed the logic of the hack_via_ws settings parameter, and added wss protocol for consistency.

ibc:
> Starting with Chrome 47, getUserMedia() requests are only allowed from secure origins

Yes, but this is irrelevant for the issue.


> When JSSIP is connected to an Asterisk server through https and wss, calls to the extension will fail (asterisk reports the extension unavailable) while it's registered, sip show peers shows status as unreachable:
1001/1001 1.0.x.x D Yes Yes 1903 UNREACHABLE

Please, open a bug in Asterisk explaining that and also explaining that `;transport=wss` does NOT exist. I won't accept hacks in JsSIP just to make it work with buggy servers for which no issue has been reported.


> Maybe it's not defined in the standards (and can be a Chrome only requirement), setting SIP URI transport parameter to 'wss' does solve the problem

Chrome knows nothing about SIP nor about what you send/receive in WebSocket messages. This is not about "Chrome and SIP".


> and the extension status becomes ok, and can be called without any further issues:
1001/1001 1.0.x.x D Yes Yes 1990 OK (395 ms)

I trust you, but as said above, I will just add hacks to JsSIP if, at least, the issue has been properly reported in the project of the buggy server (Asterisk in this case).


Regards.

orszaczky:
Further reference https://github.com/onsip/SIP.js/blob/master/src/UA.js#L1038
They probably also ran into the same issue and added 'wss' as transport parameter.

orszaczky:
Ok before I report this to Asterisk, in RFC7118 you linked above says:

5.1.  Via Transport Parameter

   Via header fields in SIP messages carry a transport protocol
   identifier.  This document defines the value "WS" to be used for
   requests over plain WebSocket connections and "WSS" for requests over
   secure WebSocket connections (in which the WebSocket connection is
   established using TLS [RFC5246] with TCP transport).

   The updated augmented BNF (Backus-Naur Form) [RFC5234] for this
   parameter is the following (the original BNF for this parameter can
   be found in [RFC3261], which was then updated by [RFC4168]):

     transport  =/  "WS" / "WSS"

is this not the same parameter we are talking about?

ibc:
2016-02-12 15:44 GMT+01:00 Gabor Orszaczky <notifications@github.com>:

>  transport  =/  "WS" / "WSS"
>
> is this not the same parameter we are talking about?
>

That refers to the transport in the Via header:

  Via: SIP/2.0/WSS


Yes, it is rare, but when using  SIP over TLS the SIP URI in the Contact
has ;transport=tcp and the Via is SIP/2.0/TLS, and when using SIP over WSS
the SIP URI in the Contact has ;transport=ws and the Via is SIP/2.0/WSS.

Ugly, but that's the correct and only valid option.



-- 
Iñaki Baz Castillo
<ibc@aliax.net>

nsnake:
I alos has this problem  and SIPJS had fixed it,

jmillan:
This will be doable in 1.0.0 without the need of a hacky option.

ibc:
> I alos has this problem and SIPJS had fixed it

@nsnake have you reported the bug to your **SIP server** vendor? this is not a but in JsSIP so JsSIP has **nothing** to fix.

62mkv:
I have reported this in Asterisk JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-25840

jmillan:
Doesn't apply anymore with the JsSIP.Socket interface

PR #371, opened 2016-04-27
sdp replacements: new RTCSession API methods to replace the SDP

jmillan:
These methods may be defined by the user in order to modify the given SDP.

- receivedSDP({type:type, sdp:sdp}) is executed before passing a remote
  SDP to RTCPeerConnection.

- sendingSDP({type:type, sdp:sdp}) is executed before sending a local SDP.


jmillan:
@ibc, @saghul,

Can you have a quick look at this MR?

The description speaks by itself.

saghul:
Belated LGTM

PR #374, opened 2016-05-25
JsSIP Socket interface

jmillan:
Make JsSIP socket agnostic.

The new JsSIP.Socket interface abstracts JsSIP from the mechanism used to send
and receive SIP traffic. JsSIP internal transport deals now with this interface
and hence, it is not attached to use the built-in WebSocket.

A new configuration parameter 'sockets' allows the user to provide it's own
sockets, which must implement the JsSIP.Socket interface.

The internal transport management and recovery handling has been
rewriten, leveraging the UA from any transport logic.

'hack_via_tcp' and 'hack_via_ws' configuration parameters have been removed.
The Via transport parameter is now seteable per Socket (see via_transport
Socket parameter).

### DIFF
--- a/lib/Socket.js
+++ b/lib/Socket.js
@@ -1,0 +1,62 @@
+module.exports = Socket;
+
+/**
+ * Interface documentation: http://jssip.net/documentation/$last_version/api/socket/
+ *
+ * interface Socket {
+ *  attribute String via_transport
+ *  attribute String url
+ *  attribute String sip_uri
+ *
+ *  method connect();
+ *  method disconnect();
+ *  method send(data);
+ *
+ *  attribute EventHandler onconnect
+ *  attribute EventHandler ondisconnect
+ *  attribute EventHandler ondata
+ * }
+ *
+ */
+
+
+/**
+ * Dependencies.
+ */
+var Utils =       require('./Utils');
+var debugerror =  require('debug')('JsSIP:ERROR:Socket');
+
+function Socket() {}
+
+Socket.isSocket = function(socket) {
+  if (typeof socket === 'undefined') {
+    debugerror('Undefined JsSIP.Socket instance');
+    return false;
+  }
+
+  // Check Properties
+  try {
+    ['via_transport', 'url', 'sip_uri'].forEach(function(prop) {
+      if (!Utils.isString(socket[prop])) {
+        debugerror('Missing or invalid JsSIP.Socket property: '+ prop);
+        throw new Error();
+      }
+    });
+  } catch(e) {
+    return false;
+  }
+
+  // Check Methods
+  try {
+    ['connect', 'disconnect', 'send'].forEach(function(method) {
+      if (!Utils.isFunction(socket[method])) {
+        debugerror('Missing or invalid JsSIP.Socket method: '+ method);
+        throw new Error();
+      }
+    });
+  } catch(e) {
+    return false;
+  }
+
+  return true;
+};
--- a/lib/WebSocket.js
+++ b/lib/WebSocket.js
@@ -1,0 +1,144 @@
+module.exports = WebSocket;
+
+/**
+ * Dependencies.
+ */
+var Grammar =     require('./Grammar');
+var debug =       require('debug')('JsSIP:WebSocket');
+var debugerror =  require('debug')('JsSIP:ERROR:WebSocket');
+
+// 'websocket' module uses the native WebSocket interface when bundled to run
+// in a browser.
+var W3CWebSocket = require('websocket').w3cwebsocket;
+
+function WebSocket(url, node_ws_options) {
+  debug('new()');
+
+  var sip_uri = null;
+  var via_transport = null;
+
+  this.ws = null;
+
+  this.node_ws_options = node_ws_options || {};
+
+  // setting the 'scheme' alters the sip_uri too (used in SIP Route header field)
+  Object.defineProperties(this, {
+    via_transport: {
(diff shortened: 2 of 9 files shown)


lib/Socket.js
```diff
@@ -0,0 +1,62 @@
...
+    debugerror('Undefined JsSIP.Socket instance');
+    return false;
+  }
+
+  // Check Properties
+  try {
+    ['via_transport', 'url', 'sip_uri'].forEach(function(prop) {
+      if (!Utils.isString(socket[prop])) {
+        debugerror('Missing or invalid JsSIP.Socket property: '+ prop);
+        throw new Error();
```
saghul: why throw and catch here instead of just returning false?

lib/Socket.js
```diff
@@ -0,0 +1,62 @@
...
+  } catch(e) {
+    return false;
+  }
+
+  // Check Methods
+  try {
+    ['connect', 'disconnect', 'send'].forEach(function(method) {
+      if (!Utils.isFunction(socket[method])) {
+        debugerror('Missing or invalid JsSIP.Socket method: '+ method);
+        throw new Error();
```
saghul: ditto

lib/WebSocket.js
```diff
@@ -0,0 +1,144 @@
...
+
+function WebSocket(url, node_ws_options) {
+  debug('new()');
+
+  var sip_uri = null;
+  var via_transport = null;
+
+  this.ws = null;
+
+  this.node_ws_options = node_ws_options || {};
```
saghul: Object.assign({}, node_ws_options) ?

jmillan:
@ibc, @saghul, consider giving it a quick look. The description speaks by itself. Documentation changes are already uploaded:

http://jssip.net/documentation/1.0.x/api/socket/
http://jssip.net/documentation/1.0.x/api/ua_configuration_parameters/#parameter_sockets

lib/Socket.js
jmillan: `return` inside a `foreEach` just continues on the next iteration

lib/Socket.js
jmillan: ditto :-)

saghul:
I had a quick (not too deep) look at the code. This is great Jose!

Now that we are using semver, this would make it a 2.0.0, since we break backwards compatibility. But I like the change!

lib/WebSocket.js
jmillan: Nice :-). Just something to be applied gloally in a separate commit IMHO.

lib/Socket.js
saghul: oh, I had missed we were inside the `forEach`. Sorry for the noise!

jmillan:
It doesn't break compatibility :-)

jmillan:
BTW, fastest review ever 🍰

lib/WebSocket.js
saghul: Well, we can write new code The New Way (TM), right? Then adjust the rest. Gradual changes are always easier to make than gigantic ones.

jmillan:
> It doesn't break compatibility :-)

Well, it does. Since hack_via_[tcp,udp] config parameters have been deprecated

jmillan:
I perfectly agree on making it 2.0.0.

Lets wait for any comment from @ibc

ibc:
@saghul `Object.assign()` is a ES6 feature not present in many browsers. We don't need to deal with that for now.

ibc:
Good work @jmillan, some concerns coming to my mind:

* Does this break API so we would need to release 2.0.0?

* We could entirely remove the Node.js `websocket` module (which would avoid a native add-on building every time JsSIP is installed via `npm`) and, instead, provide a separate `jssip-node-websocket` module providing a `Socket` implementation just for Node.js.  

* May you provide a basic usage example (in browser with native WebSocket)? It's not clear for me whether the user needs to instantiate a WebSocket `Socket` instance or not in this scenario.

ibc:
> `hack_via_[tcp,udp]` config parameters have been deprecated

Deprecated? or entirely removed?

jmillan:
>>hack_via_[tcp,udp] config parameters have been deprecated

>Deprecated? or entirely removed?

Completely removed

saghul:
> Completely removed

❤️

saghul:
> Does this break API so we would need to release 2.0.0?
> 

Yep.

> We could entirely remove the Node.js websocket module (which would avoid a native add-on building every time JsSIP is installed via npm) and, instead, provide a separate jssip-node-websocket module providing a Socket implementation just for Node.js.
> 

Sounds good to me!

jmillan:
> We could entirely remove the Node.js websocket module (which would avoid a native add-on building every time JsSIP is installed via npm) and, instead, provide a separate jssip-node-websocket module providing a Socket implementation just for Node.js.

Agreed. Let me check it a bit later 👍

ibc:
@jmillan as I'm familiar with the node `websocket` module and its integration in JsSIP I may remove it stuff later once this PR is merged.

jmillan:
Agreed @ibc,

The branch has been merged

ibc:
yeah

pjkundert:
I'd like to find / implement a plain TCP Transport for JsSIP.  I'm pretty sure I can't be the first one to want this, but I can't find any indication that anyone has done it!

If you know of any implementations of JsSIPP.Socket via plain TCP, could you let me know?  Thanks,

PR #380, opened 2016-06-17
Spelling fix in debug line

davies147:
sesssion becomes session

PR #381, opened 2016-06-17
Clean up trailing whitespace in built lib/Grammar.js

davies147:
lib/Grammar.js is built by PEGJS, but it leaves lots of trailing whitespace on lines.

I thought it would be nice to clean that up.


ibc:
Unfortunately this commit is already included in your #380, so you can close this PR.

davies147:
Yes I messed up and accidentally added it to the previous branch - my error. Will close this one.

PR #384, opened 2016-06-24
choose UPDATE/RE-INVITE on sending session refresh request

michelepra:
as RFC 4028 refresher can send re-INVITE or UPDATE.
Introduce JsSIP SIP
User Agent parameters (session_timers_use_update) to enable/disable
UPDATE in session refresh request


ibc:
There is a good reason for usunge UPDATE for session timers rather then reINVITE:

The purpose of session timers is to keep the **signaling** open and be able to check it. Using UPDATE for that is perfect. In the other side, using reINVITE means renegotiating the SDP session (because the INVITE/200/ACK must handle a SDP Offer/Answer) which is 100% unnecessary if the only we need is to check the session status.

Also, your pull request makes reINVITE the default session timers mechanism, which changes the current default settings.

I'm sorry but I strongly consider that using UPDATE for session timers is the proper way to go and, given that we are in 2016, I hope SIP servers, proxies and endpoints MUST implement the UPDATE method (defined in RFC 3311 in 2002, 14 years ago).

michelepra:
I agree with you that is better of use UPDATE rather reINVITE but the purpose of [RFC](https://en.wikipedia.org/wiki/Request_for_Comments) is to define an Internet Standard, and in software engineering follow standard was the goal for every developer.

If you write that jsSIP support RFC 4028 this mean that i dont trouble to check that all thing of this RFC was developed (or you must specify what is or not is present).
If, for some reason, i need reINVITE i hope to use this by valorizing some parameter.

>Also, your pull request makes reINVITE the default session timers mechanism, which changes the current default settings.

Absolutely no, check again the code, new parameter name was 

> session_timers_use_update

 and default value is true. This mean that default session timers mechanism remaining UPDATE

ibc:
OK, let us recheck it during this week.

ibc:
@jmillan any thoughts on this?

jmillan:
I'm ok with this PR. 👍

lib/RTCSession.js
```diff
@@ -2672,8 +2673,9 @@ function handleSessionTimersInIncomingResponse(response) {
 }
 
 function runSessionTimer() {
-  var self = this;
-  var expires = this.sessionTimers.currentExpires;
+  var self = this,
+	expires = this.sessionTimers.currentExpires,
+	eventHandlers;
```
ibc: Please, put each variable declaration in a single line `var xxx = xxx;`.

lib/RTCSession.js
```diff
@@ -2683,16 +2685,23 @@ function runSessionTimer() {
   if (this.sessionTimers.refresher) {
     this.sessionTimers.timer = setTimeout(function() {
       if (self.status === C.STATUS_TERMINATED) { return; }
+	  eventHandlers = {
+		  succeeded: function (response) {
+			  handleSessionTimersInIncomingResponse.call(self, response);
+		  }
+	  };
```
ibc: Here there is a mix of tabs and spaces. Please, use two spaces for indentation.

lib/RTCSession.js
```diff
@@ -2683,16 +2685,23 @@ function runSessionTimer() {
...
-      });
+      if (self.sessionTimers.useUpdate) {
+          sendUpdate.call(self, {
+			eventHandlers: eventHandlers  
+		  });
+      } else {
+          sendReinvite.call(self, {
+			eventHandlers: eventHandlers  
+		  });
+      }
```
ibc: Same here, mix of tabs and spaces. Use two spaces, please.

michelepra:
Many time passed from last review, now i correct file as requested

michelepra:
+1

michelepra:
i will create new PR next week. Before i must generate new clone of master because my branch now include fix of #427 and other minor change (it would not be bad idea if is integrated into jssip master)

jmillan:
Thanks, no hurries.

We would honestly like to do the decoupling ourselves. We have now more experience than we had before and more clear ideas on how to do it.

jmillan:
@michelepra, for me it would be easy to do theport: just in case you don't want to mess up with it.

michelepra:
ok, sorry @jmillan  this week i'm extremly full, no time to work on this. If you want close this pr and do the port, for me nothing change. Very interesting thing in my branch is decoupling, that work very well in production with nodejs enviroment, but i'd like someday to use official code to do same thing. Any plan when release?

jmillan:
Landed in 3.1.0.

Thanks!

PR #388, opened 2016-06-30
Remove dangling callbacks after WebSocketInterface.disconnect().

davies147:
- These callbacks may be delayed waiting for the actual close,
- which interferes with the smooth running of any replacement
- WS, so clear the callbacks and fake the close event instead.

The use-case which required this was to allow a controlled UA.stop() on loss of network and UA.start() when network returned. If this happend quickly, the callback events from the UA.stop() would cause the newly started instance to fail.

### DIFF
--- a/lib/WebSocketInterface.js
+++ b/lib/WebSocketInterface.js
@@ -54,7 +54,7 @@ WebSocketInterface.prototype.connect = function () {
   }
 
   if (this.ws) {
-    this.ws.close();
+    this.disconnect();
   }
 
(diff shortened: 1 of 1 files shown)


jmillan:
Hi @davies147,

I've pushed a commit in master which unbinds the WS event callbacks on disconnecting the socket, as you well indicated, plus resets the internal `UA.stop()` grace timeout of 2 seconds that was making impossible to achieve an immediate UA restart.

Thanks a lot!

PR #405, opened 2016-11-02
Prevent Parser from throwing a TypeError

### LINKED ISSUE #404: Parser is setting 'parsed' variable to undefined, then trying to find it's length.
I've recently noticed a few errors coming from the Parser module.

```TypeError: Cannot read property 'length' of undefined```

It appears that [these](https://github.com/versatica/JsSIP/blob/master/lib/Parser.js#L94-L98) are the offending lines for most of these cases, however the same issue is repeated a few times in that module. The issue is as follows:

```
if (parsed === -1) {
  parsed = undefined;
}

length = parsed.length;
```

If `parsed` is set to `-1`, then it will be reassigned to `undefined`, and so will throw the above error when trying to read it's length.

I'd be happy to submit a PR to fix this issue if that would be beneficial.

joshfarrant:
Prevent Parser from throwing a TypeError when 'parsed' variable is equal to '-1'

Fixes #404


ibc:
LGTM. @jmillan?

ibc:
Thanks @joshfarrant. May you please fill the [CLA form](http://jssip.net/contribute/) so we can incorporate this PR into JsSIP?

joshfarrant:
Done 🙂

ibc:
Merged. Thanks.

PR #408, opened 2016-11-14
Remove rtcninja dependency

ibc:
And use webrtc-adapter instead.


saghul:
:cake:

ibc:
Done via console.

jmillan:
👍

PR #423, opened 2017-02-09
Add before_reregister option.

michaelstephendavies:
Set this to a function with a single callback argument.
Before JsSIP reregisters it will call the function.
The reregister will happen once you call the callback.
This gives you a chance to change some settings which
will affect the registration (e.g. generate a time-limited
token to register with).

### DIFF
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -146,7 +146,14 @@ Registrator.prototype = {
           // For that, decrease the expires value. ie: 3 seconds
           this.registrationTimer = setTimeout(function() {
             self.registrationTimer = null;
-            self.register();
+
+            if (self.ua.configuration.before_reregister) {
+              self.ua.configuration.before_reregister.call(self.ua, function() {
+                self.register();
+              });
+            } else {
+              self.register();
+            }
           }, (expires * 1000) - 3000);
 
           //Save gruu values
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -698,6 +698,7 @@ UA.prototype.loadConfig = function(configuration) {
     register_expires: 600,
     register: true,
     registrar_server: null,
+    before_reregister: null,
 
     use_preloaded_route: false,
 
@@ -876,6 +877,9 @@ UA.prototype.loadConfig = function(configuration) {
       case 'ha1':
         debug('- ' + parameter + ': ' + 'NOT SHOWN');
         break;
+      case 'before_reregister':
+        debug('- ' + parameter + ': ' + typeof settings[parameter] === 'function' ? '<function>' : JSON.stringify(settings[parameter]));
+        break;
       default:
         debug('- ' + parameter + ': ' + JSON.stringify(settings[parameter]));
     }
@@ -913,6 +917,7 @@ UA.configuration_skeleton = (function() {
       'registrar_server',
       'sockets',
       'use_preloaded_route',
+      'before_reregister',
 
       // Post-configuration generated parameters
       'via_core_value',
@@ -1120,6 +1125,12 @@ UA.configuration_check = {
     use_preloaded_route: function(use_preloaded_route) {
       if (typeof use_preloaded_route === 'boolean') {
         return use_preloaded_route;
+      }
+    },
+
+    before_reregister: function(before_reregister) {
+      if (typeof before_reregister === 'function') {
+        return before_reregister;
       }
     }
   }


jmillan:
Sorry, accepting a `function` as a configuration parameter in not an option.

We already indicated that this callback is not required at all.

ibc:
Agreed with @jmillan. You can take the [Registrator](http://jssip.net/documentation/3.0.x/api/registrator/) instance at any time and add headers or params:

http://jssip.net/documentation/3.0.x/api/ua/#method_registrator

jmillan:
Hi,

Considering we are getting few requests on this topic, we may provide a new event which fires `on register timeout`, prior to sending the next request.

jmillan:
The new REGISTER would be sent after getting back from the new event callback unless the user decided to unregister. The user would have the chance to re-configure the Registrator within the callback. It's up to the user to make that operation fast enough so it doesn't get unavailable for much time. Or it could just use this new event to unregister,  asynchronously retrieve whatever info it needs, and once having re-configured the Registrator, register the UA again.

ibc:
What about something like this?:

```js
ua.on('willregister', (exec) => {
  getNewTokenFromMyServer()
    .then((token) => {
      ua.registrator.setStuff(token);
      exec();
  });
});
```

ibc:
I mean, if the app subscribes to the `willregister` event, then JsSIP won't automatically send a re-REGISTER but, instead, will rely on the `exec` method given in the `willregister` event be called by the app.

jmillan:
I'm Ok with it 👍

ibc:
NOTE: Actually JsSIP already relies on whether the app set some events or not to behave differently. For example, if the app does not set a `newRTCSession` event, incoming calls would be automatically rejected. In UA.js:

```js
  if(method === JsSIP_C.OPTIONS) {
    request.reply(200);
  } else if (method === JsSIP_C.MESSAGE) {
    if (this.listeners('newMessage').length === 0) {
      request.reply(405);
      return;
    }
    message = new Message(this);
    message.init_incoming(request);
  } else if (method === JsSIP_C.INVITE) {
    // Initial INVITE
    if(!request.to_tag && this.listeners('newRTCSession').length === 0) {
      request.reply(405);
      return;
    }
  }
```

jmillan:
I would definitely call it `registertimeout`

jmillan:
Hi @danjenkins,

`registerTimeout` (yes, camelCase) does express exactly what it is notifying. This event should be triggered just when the timer [expires](https://github.com/versatica/JsSIP/blob/master/lib/Registrator.js#L149).

It is user's choice to block all registers upon such event or not, but the trigger of the event is the register timeout.

ibc:
JsSIP sends a re-REGISTER some seconds before it really needs, so IMHO it's perfectly safe to name such an event "willregister" or "registerPending" or things like that. The app then has about 10 seconds to call the given `exec()` method to trigger a re-REGISTER.

ibc:
What about "needRegistration"?

ibc:
Assuming "registerTimeout" event, why would we "also" need a "registerTimeout" event?

ibc:
Sorry, comment updated

jmillan:
Sorry, I had to leave for a while.

Maybe I was wrong and `registerTimeout` event does not provide any benefit on this scenario.  `registerPending`  event listener could be responsible for running `UA.register()`. If no `registerPending` listener is present then the Registrator would not fire the event and would send the REGISTER request as done now.

This way there's no need to provide the `exec` method in the event callback for the user to run it.

ibc:
@jmillan I don't follow you. If no `exec` method is provided within the event, then we have no way to delay the new registration refresh and, as @danjenkins explained, async operations may be needed to set something in the new REGISTER (before it's sent).

ibc:
@danjenkins you mention both `registerPending` (comes with `exec`) and `registerTimeout`, but in your second paragraph you tell about a single `registerTimeout` with `exec`. Do you mean `registerPending`? or is this just about having a single event (with `exec`) and no matter it's called in some way or another?

ibc:
Sorry @jmillan, now I get you: Instead of giving a `exec` method, the app must call `ua.register()` (which in fact would be exactly what `exec()` would do, am I right?)

jmillan:
> Sorry @jmillan, now I get you: Instead of giving a exec method, the app must call ua.register() (which in fact would be exactly what exec() would do, am I right?)

That's it.

jmillan:
I'm Ok with it. @ibc?

ibc:
100% fine with me.

ibc:
Yep. Just let me reopen the event name subject. `registerTimeout` seems that something bad has **already** happened. What about `registerWillTimeout` or `registrationNeeded`? Just wondering, I'm open to whichever.

ibc:
I don't like `registrationEnding` much... `registrationExpires`?

ibc:
Works For Me ®

jmillan:
👍

ibc:
Definitely yes, like it. And we break nothing doing that.

ibc:
Looks nice.

> Where is the right place to add docs?

https://github.com/versatica/jssip.net

But don't do it. The project is not documented. We'll add it when the PR is merged.

ibc:
Thanks!

PR #426, opened 2017-02-20
Do not overwrite eventHandlers in sendDTMF function

AlphaBetaPeter:
In the current implementation (2.0.x and up) all eventhandlers are overwritten in the sendDTMF function. 

In addition to this the eventhandlers 'succeeded' and 'failed' (like stated in the documentation) are never called because the actual implementation requires them to be called 'onSuccessResponse' and 'onErrorResponse'.

Regards,
Peter

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -919,7 +919,7 @@ RTCSession.prototype.sendDTMF = function(tones, options) {
       var dtmf = new RTCSession_DTMF(self);
       options.eventHandlers.failed = function() {
         self.tones = null;
-      }
+      };
       dtmf.send(tone, options);
       timeout = duration + interToneGap;
     }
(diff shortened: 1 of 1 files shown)


jmillan:
Hi @AlphaBetaPeter,

Thanks for reporting the issue. 

There is a problem in the documentation, and the implementation does not correspond to it. It'll be fixed as soon as possible.

jmillan:
Hi @AlphaBetaPeter,

The 'eventHandlers' argument in `RTCSessioin::sendDTMF` has been removed, since it should never have been there.

You can set the DTMF event callbacks on `RTCSession::newDTMF` event handler, on the `dtmf` instance.

This is ready in new release 3.0.3.

Kind regards.

jmillan:
[Documentation](http://jssip.net/documentation/3.0.x/api/session/#method_sendDTMF) has also been updated:

PR #441, opened 2017-05-22
Add in an event when a registration is expiring

scgm11:


scgm11:
sorry was checking at the changes and click by error

lib/Registrator.js
```diff
@@ -142,11 +142,17 @@ Registrator.prototype = {
...
             self.registrationTimer = null;
-            self.register();
+            // If there are no listeners for registrationExpiring, renew registration
+            // If there are listeners, let the function listening do the register call
+            if (self.ua.listeners('registrationExpiring').length === 0) {
+              self.register();
+            } else {
+              self.ua.emit('registrationEnding');
+            }
           }, (expires * 1000) - 3000);
```
ibc: Could you please increase those 3000 ms to something higher such as 10000ms? 3 seconds is a bit risky if async operations would be performed before re-registering.

PR #442, opened 2017-05-23
Registration expiring event

danjenkins:
This allows people to listen for a registration expire event and then to manually call ua.register() themselves after doing some fancy logic 

This re-implements what the aim of #423 was attempting to fix.

### DIFF
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -153,7 +153,7 @@ Registrator.prototype = {
             } else {
               self.ua.emit('registrationExpiring');
             }
-          }, (expires * 1000) - 3000);
+          }, (expires * 1000) - 5000);
 
           //Save gruu values
           if (contact.hasParam('temp-gruu')) {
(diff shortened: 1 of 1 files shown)


ibc:
Thanks @danjenkins 
May you please accomplish with the little required change?

ibc:
Also, may you please fill the [CLA](http://jssip.net/contribute/)?

ibc:
IMHO nobody should rely on the fact that JsSIP re-registers 3 seconds before the negotiated "expires".

Regarding making that configurable, I think it wouldn't be a very "cool" option and I prefer to leave it in a fixed way that "just works" (as now).

Under your opinion, is 3 seconds enough for common REST/API/HTTP/whatever request/response?

I may propose increase it to just 5 seconds.

ibc:
Cool, can you please commit it? I've already have JsSIP project open to publish a new version XD

ibc:
3.0.12 is [out](https://github.com/versatica/JsSIP/blob/master/CHANGELOG.md) and new event [documented](http://jssip.net/documentation/3.0.x/api/ua/#event_registrationExpiring) :)

Thanks @danjenkins !

PR #448, opened 2017-07-11
Implements rtpTimeout and byeReceived RTCSession events

nicketson:
If these events have handlers registered, call termination is
postponed. For both events, the event handler receives an event object
that has a 'finish' callback function. The callback function takes a
single argument 'extraHeaders'.

For the rtpTimeout event, the SIP 408 terminate message will be sent
with whatever extra headers are passed to the callback function.

For the byeReceived event, the SIP 200 response to the BYE message
will be sent with whatever extra headers are passed to the callback
function.

We are using these new callbacks so that we can send call
quality/statistics data via extra SIP headers.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1236,6 +1236,11 @@ RTCSession.prototype.receiveRequest = function(request) {
   var contentType,
       self = this;
 
+  function respondToBye(extraHeaders) {
+    request.reply(200, null, extraHeaders);
+    ended.call(self, 'remote', request, JsSIP_C.causes.BYE);
+  }
+
   if(request.method === JsSIP_C.CANCEL) {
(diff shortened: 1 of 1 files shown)


jmillan:
Hi,

I would just pass the extra headers array in the event object callback so the user just edits the array if she wants, avoiding the user to execute any method that sends the reply.

nicketson:
For my use case, my rtpTimeout and byeReceived event handlers execute RTCPeerConnection.getStats(). getStats is asynchronous so I need to delay sending the reply until the getStats promise resolves. 

> I would just pass the extra headers array in the event object callback so the user just edits the array if she wants, avoiding the user to execute any method that sends the reply

Perhaps this can be implemented in addition to the callback so that the event can be handled either synchronously or asynchronously?

jmillan:
Hi, I'll push a generic solution for this. Thanks for bringing it up @nicketson

PR #450, opened 2017-07-12
Use promise chaining to prevent PeerConnection state race-conditions

davies147:
The change which stopped using 'pranswer' introduced a race condition.

If a 183+SDP arrives, and the setRemoteDescription() is called, and then a 200+SDP arrives VERY quickly, and before the previous setRemoteDescription() has completed and resulted in a 'stable' state, then the attempt to apply the 2nd SDP causes a call drop due to invalid PC state.

I have tried to implement a general-purpose solution where all Promise capable PeerConnection calls are chained. It solves my issue.

PS. I am not keen on my 'connectionPromise' variable name - Feel free to find a better name!

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -65,6 +65,9 @@ module.exports = class RTCSession extends EventEmitter
     // The RTCPeerConnection instance (public attribute).
     this._connection = null;
 
+    // Prevent races on serial PeerConnction operations
+    this._connectionPromiseQueue = Promise.resolve();
+
     // Incoming/Outgoing request being currently processed.
     this._request = null;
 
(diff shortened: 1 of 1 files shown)


jmillan:
Hi @davies147,

Thanks for reporting this and sorry for the delay on the feedback. Your approach is very nice.

We are going to check this carefully prior to proceed with the implementation. We are a bit busy right now but will come back ASAP.

davies147:
No worries on the delayed response - I do like a Promise to solve a problem ;-)

If it helps, I've been running this on several tens of endpoints since I posted the pull request, and it has not bitten me *yet*

jmillan:
Hi @davies147,

Would you fill the project [CLA](http://jssip.net/contribute/)?

We've recently transpiled the code to ES6, so your patch will no longer merge. You can create a new PR or we can do it for you otherwise.

BTW: I do prefer the name `connectionPromiseQueue`

davies147:
CLA completed as requested.

Your suggested variable name is an improvement - I struggled to find a good
name :)

I will look and see how easily it can be merged to your ES6 version, but I
will not complain if you have already done it either. Git and co-operation
is what it should be like.

Regards,
Steve

jmillan:
Thanks @davies147,

Don't worry, I'll do it for you :-)

davies147:
It is nearly done...

davies147:
I have pushed the edit, and run dist and tests against it but have not yet run it live.

jmillan:
Branch merged.

Thanks @davies147!

PR #466, opened 2017-10-23
fix for utf8Encode

MaxKirillov:
Hi,
I found a problem in Utils.js file. Please see diff.


jmillan:
Thanks @MaxKirillov,

It has been fixed along with another typo.

PR #471, opened 2017-11-02
Fixed undefined this

iclems:
When called within setTimeout, the context for userMediaSucceeded was wrong.


jmillan:
Thanks,

May you please add en entry to the JsSIP [CLA](http://jssip.net/contribute/)?

Also, please, may you modify your PR so only the change into the source code is done (in RTCSession.js)?

We'll generate the jssip.js and the rest of changes as part of the release.

Thanks.

jmillan:
Couldn't live with a well known bug.

Fix pushed on newest release.

Thanks for reporting!

PR #474, opened 2017-11-09
UA.isRegistered() always returned false

pejsajan:
it is broken after ES6 rewrite (since 3.0.19 version)

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -206,7 +206,7 @@ module.exports = class UA extends EventEmitter
    */
   isRegistered()
   {
-    if (this._registrator.registered)
+    if (this._registrator._registered)
     {
       return true;
     }


jmillan:
Ups, already fixed in last revision.

Thanks

pejsajan:
Hi,

in latest version (3.0.27) there were added registered() getter into Registrator, but method isRegistered() in UA was not fixed - or am I wrong?

jmillan:
No need to touch UA.js. The issue was that `Registrator` was missing the getter for `registered`

````js
      /**
       * Registration state.
       */
      isRegistered()
      {
      if (this._registrator.registered)
        {
          return true;
        }
        else
        {
          return false;
        }
      }
`````

pejsajan:
ok

PR #479, opened 2017-11-21
Emitting sdp event before creating offer / answer

douglaseel:
The sdp event is called after the creation of the offer and answer in two points of the code. This pull request corrects these points, emitting sdp event before.


ibc:
Please don't include bundled (dist) files in the PR but just the exact changes to files in `lib/`.

ibc:
What do you think @jmillan ?
Note: if ok, let's squash it rather than merging.

jmillan:
Thanks @douglaseel,

We were doing it this way in some places but not everywhere.

I've cleaned up the commit and released a new version.

PR #485, opened 2017-12-28
Upgraded the vinyl versions to fix the 'premature close' in OSX

garysoed:
See [a similar issue](https://github.com/versatica/mediasoup-demo/issues/15).


ibc:
Thanks @garysoed 
This is fixed in 268be95aaf7b0187baa00d8eb764e9cd6b61401b. I've also upgraded all he dependencies in that commit.

PR #486, opened 2017-12-28
Allow client of JsSIP to specify custom User Agent

garysoed:
Our SIP server uses the User-Agent for identifying the client. Since there are potentially many clients using JsSIP, identifying the client as "JsSIP" is insufficient.

Let me know if there are better ways to implement this!


ibc:
Just curious: what do you mean with "identify the client based on the User-Agent"?

garysoed:
For example, suppose we have a SIP server that many companies want to use. Apple, Google, Microsoft all use JsSIP to talk to this SIP server. But the server wants to know if a SIP message comes from Apple / Google / Microsoft (probably for resource allocation?). With the current implementation, Apple / Google / Microsoft all sends "JsSIP" as the User Agent, so the server won't be able to differentiate between them.

Hope that makes sense. Thanks!

ibc:
I assume that all those JsSIP servers are under your control, right? If so, why don't you just add a custom SIP header into REGISTER and INVITE requests?

garysoed:
Unfortunately, I don't have control over the SIP server. The SIP server currently looks at the User-Agent to determine the client information.

ibc:
May you rename `client_user_agent` to just `user_agent` and remove `dist/jssip.min.js` from the PR, please?

ibc:
We are coming back to the real life. This and other issues and PRs will be managed soon. Thanks.

garysoed:
Done. I've removed `dist/jssip.min.js` from the PR Why not `dist/jssip.js` as well btw?
I've also renamed `client_user_agent` to `user_agent`.

jmillan:
Squashed and merged. 

Thanks!

PR #488, opened 2017-12-30
Fixed session.sendInfo()

lrorthmann:
Calling the session.sendRequest() with the correct parameters, so it can send the correct SIP INFO message.


ibc:
Upps, @jmillan may be a bug in the migration to ES6?

jmillan:
Merged,

Thanks!

PR #496, opened 2018-01-25
extend DTMF support for interfacing with a wider range of networks

emiliorcueto:
The option of supporting RFC2833/4733 and/or inband DTMF should be strongly considered.  Currently, global providers such as Twilio still do not support DTMF via SIP INFO.  Also, as in our use case, we need to support interfacing with legacy networks which require a user's interaction via IVR.

As it stands here, this pull request leaves the default functionality of DTMF over SIP INFO as is.  The transport can be changed however by setting a new UA config: `session_dtmf_transport` to either "RFC2388" or "IN_BAND".  

_Note: Although the option for RFC2833 now exists, it's functionality must still be implemented.  Only inband DTMF functionality within the audio stream is implemented in this pull request_


lib/RTCSession.js
```diff
@@ -1009,13 +1009,28 @@ module.exports = class RTCSession extends EventEmitter
...
+        // send DTMF according to transport config
+        switch (this._ua.config.session_dtmf_transport.toUpperCase())
+        {
+          case JsSIP_C.DTMF_TRANSPORT.RFC2833:
+            // TODO: support payload 101, et al.
+            break;
+
+          case JsSIP_C.DTMF_TRANSPORT.IN_BAND:
+            // send DTMF in current audio RTP stream
+            this._getDTMFRTPSender().insertDTMF(tone, duration, interToneGap);
```
jmillan: This will throw everytime `_getDTMFRTPSender()` doesn't return a  `RTCDTMFSender` instance.

Also it is not notifying the user about the new sent DTMF, as [we do](http://jssip.net/documentation/3.0.x/api/session/#event_newDTMF) with the INFO DTMFs.

If `sendDTMF()` is called before a previous call to this method has finished sending all the tones, the remaining ones will be discarded and not sent.

jmillan:
Hi @emiliorcueto,

I've commented on your code. Overall this feature belongs to `RTCSession` only. It is the only file that should be changed.

A new option on `RTCSession::sendDTMF()` would suffice. Ie: (`type: INFO|INBAND)`

The additions on `RTCSession::sendDTMF()` are not properly done. (details on inline comments ).

I'd suggest you to check the [RTCDTMFSender specification](https://www.w3.org/TR/webrtc/#peer-to-peer-dtmf).

jmillan:
I'm closing this PR due to inactivity

PR #502, opened 2018-02-13
Fix ordering of newRTCSession and peerconnection events

StarLeafRob:
The peer connection must be created after the newRTCSession method is called, otherwise the "peerconnection" event is fired before the "newRTCSession" event. This would be bad, because one typically attaches listeners for the "peerconnection" event in the "newRTCSession" callback!

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -357,9 +357,6 @@ module.exports = class RTCSession extends EventEmitter
 
     this._id = this._request.call_id + this._from_tag;
 
-    // Create a new RTCPeerConnection instance.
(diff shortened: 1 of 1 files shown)


StarLeafRob:
1) Apologies for the exclamation mark, I didn't mean it to be rude.
2) That's true - it's just a bit confusing if there's an event that can be
fired before the user has a chance to attach a listener for it. The
connection is also always present after calling RTCSession:answer - perhaps
the RTCSession:peerconnection event is unnecessary then, and should be
removed?

jmillan:
Yes, very probably we can get rid of it. Let us check it.

Thanks a lot for the comment, indeed!

PR #505, opened 2018-03-03
RTCSession: prefer promises over callbacks for readability

jmillan:
This is a refactor exercise on replacing callbacks with promises.

It provides readability and makes it easier to follow the flow.

The approach has been dealing with each scope in its own block. Ie:

```js
return Promise.resolve()
  // Block A
  then(() =>
  {
     return MethodReturningPromise
       .catch(() =>
       {
          // Block A error handling here.
          return Promise.reject('Block A');
       });
  })
  // Block B
  then(() =>
  {
     return MethodReturningPromise
       .catch(() =>
       {
          // Block B error handling here.
          return Promise.reject('Block B');
       });
  })
  catch((step) =>
  {
     if (step !== 'Block A' && step !== 'Block B')
       throw new Error(Unhandled exeption);
  }
```

As said, it is just a refactor exercise that makes the code easier to follow and more readable.

I can think of adding async/await in future for less boilerplate code an more readability.., but for now it is just what it is.

It's been tested against tryit.jssip.net successfully.

would you give it a quick review whenever you have some time @ibc?

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -656,7 +656,7 @@ module.exports = class RTCSession extends EventEmitter
 
               this.emit('getusermediafailed', error);
 
-              return Promise.reject('getUserMedia');
+              throw new Error('getUserMedia');
             });
         }
       })
@@ -665,7 +665,7 @@ module.exports = class RTCSession extends EventEmitter
       {
         if (this._status === C.STATUS_TERMINATED)
         {
-          return Promise.reject(new Error('terminated'));
+          throw new Error('terminated');
         }
 
         this._localMediaStream = stream;
@@ -701,7 +701,7 @@ module.exports = class RTCSession extends EventEmitter
 
             this.emit('peerconnection:setremotedescriptionfailed', error);
 
-            return Promise.reject('setRemoteDescription');
+            throw new Error('setRemoteDescription');
           });
 
         return this._connectionPromiseQueue;
@@ -711,7 +711,7 @@ module.exports = class RTCSession extends EventEmitter
       {
         if (this._status === C.STATUS_TERMINATED)
         {
-          return Promise.reject('terminated');
+          throw new Error('terminated');
         }
 
         // TODO: Is this event already useful?
@@ -724,7 +724,7 @@ module.exports = class RTCSession extends EventEmitter
             {
               request.reply(500);
 
-              return Promise.reject('createLocalDescription');
+              throw new Error('createLocalDescription');
             });
         }
         else
@@ -734,7 +734,7 @@ module.exports = class RTCSession extends EventEmitter
             {
               request.reply(500);
 
-              return Promise.reject('createLocalDescription');
+              throw new Error('createLocalDescription');
             });
         }
       })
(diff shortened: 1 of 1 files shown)


lib/RTCSession.js
```diff
@@ -626,45 +626,53 @@ module.exports = class RTCSession extends EventEmitter
...
+          this._localMediaStreamLocallyGenerated = true;
 
-    // User media succeeded.
-    function userMediaSucceeded(stream)
-    {
-      if (this._status === C.STATUS_TERMINATED) { return; }
+          return navigator.mediaDevices.getUserMedia(mediaConstraints)
+            .catch(() =>
+            {
+              request.reply(480);
```
ibc: You are not calling `debugerror('emit "getusermediafailed" [error:%o]', error);` anymore (which is needed by the callstats.io JsSIP plugin.

lib/RTCSession.js
```diff
@@ -626,45 +626,53 @@ module.exports = class RTCSession extends EventEmitter
...
+              this._failed('local', null, JsSIP_C.causes.USER_DENIED_MEDIA_ACCESS);
 
-      this._localMediaStream = stream;
-      if (stream)
+              return Promise.reject('getUserMedia');
```
ibc: IMHO it's better to do `throw new Error('getUserMedia() failed');` here rather than returning a rejected promise.

lib/RTCSession.js
```diff
@@ -626,45 +626,53 @@ module.exports = class RTCSession extends EventEmitter
...
+        }
+      })
+      // Attach MediaStream to RTCPeerconnection.
+      .then((stream) =>
       {
-        this._connection.addStream(stream);
-      }
+        if (this._status === C.STATUS_TERMINATED)
+        {
+          return Promise.reject(new Error('terminated'));
```
ibc: Same.

lib/RTCSession.js
jmillan: Note: I reject the Promise with a `string` and not an `Error` because it is only used by the `catch()` statement on the Promise chain in order to make sure that the error has been handled by the corresponding block. It is just a label (I called it **step**) so the `catch` statement can compare it against the well known steps that may have failed (and already handled the error).

See [here](https://github.com/versatica/JsSIP/blob/promisify/lib/RTCSession.js#L760)

Having said that, the following could be more clean and generic:

```javascript
throw new Exceptions.PromiseChainError('getUserMedia');
```

```javascript
.catch((error) =>
{
  if (! (error instanceof Excepcions.PromiseChainError))
 {
   throw new Error(`Unhandled exception: ${error}`);
 }
});
```

lib/RTCSession.js
jmillan: Good catch.

jmillan:
Not ready yet. I need to recheck it.

jmillan:
Much better now.

@ibc, as you suggested I'm throwing from the errored block with a descriptive message. The `catch()` block just logs the received error.

Incoming re-INVITE and UPDATE SDP processing has been unified since they where sharing most of the code.

jmillan:
Thanks for the review @ibc.

ibc:
Hey, this is closed! should it?

jmillan:
I already merged it. Where you willing to comment more on it?

ibc:
ah, you merged it locally, ok ok :)

PR #508, opened 2018-03-21
RTCSession: custom local description trigger support

jmillan:
```
Provide a new event 'icecandidate' which allows the user to indicate
JsSIP that the local description is ready to be sent regardless of
whether the acknowledged ICE candidate is the last one notified by
RTCPeerConection or not.
```

This prevents the hack of setting a gathering timeout in JsSIP and provides more flexibility on the decision of triggering the local description.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1877,24 +1877,40 @@ module.exports = class RTCSession extends EventEmitter
         // Add 'pc.onicencandidate' event handler to resolve on last candidate.
         return new Promise((resolve) =>
         {
+          let finished = false;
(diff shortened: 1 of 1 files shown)


jmillan:
It does not break. If the event callback is not defined JsSIP behaves as before. This can only speed up the local description generation process with no side effect.

Of course, user must control when to tell JsSIP that she's fine with stop gathering ICE candidates, but that's the very facility the feature provides.

ibc:
Then it's a cool solution for a hard problem.

jmillan:
Merged

PR #519, opened 2018-05-06
Allow multiple Call-Info headers in received SIP messages

rprinz08:
According to some RFCs (e.g [RFC7852](https://tools.ietf.org/html/rfc7852), section 6.1 paragraph 6 and the example on page 45) it is possible that there are multiple Call-Info headers are present in received messages. The current implementation uses the `setHeader` method in the parser which overwrites already existing headers. The patch uses the `addHeader` method which allows multiple headers instead.


lib/Parser.js
```diff
@@ -299,6 +299,11 @@ function parseHeader(message, data, headerStart, headerEnd)
         message.event = parsed;
       }
       break;
+    // allow multiple call info headers
```
ibc: Please, keep the comments style (uppercase the first letter and end with dot).

lib/Parser.js
```diff
@@ -299,6 +299,11 @@ function parseHeader(message, data, headerStart, headerEnd)
...
+    case 'call-info':
+      message.addHeader(headerName, headerValue);
```
ibc: Why ' headerName'  instead of ' call-info' as in any other header above?

lib/Parser.js
```diff
@@ -299,6 +299,11 @@ function parseHeader(message, data, headerStart, headerEnd)
...
+      parsed = 0;
```
ibc: Why `parsed = 0;` here?

rprinz08: `parsed = 0;` to indicate a `true` function return (same way as used
in the default case block).

lib/Parser.js
rprinz08: According to RFC3261 (section 7.3.1, page 29 last paragraph on page 29)
not only the 'Call-Info' header can appear multiple times but any 
header which value can be expressed as a comma seperated list. Thus
the provided case block can be useable not only for the call-info but 
also for other headers which not need further special processing like 
in the other case blocks. Therefore I have used the `headerName` variable 
instead of the 'call-info' string literal. But can change this if you want.

lib/Parser.js
rprinz08: Will be changed

ibc:
Hey @jmillan, does this look good for you?

jmillan:
Hi @rprinz08,

I don't see any definition for `call-info` header field in [Grammar.pegs](https://github.com/rprinz08/JsSIP/blob/master/lib/Grammar.pegjs).

The header field values are being stored in `raw` in the message. 

Did you intend to do so?

Note: I'm not against to do so for header fields which are not known or relevant for JsSIP, but then there should be a generic method to do so for all  `unknown` header fields. Indeed this behaviour for all `unknown` headers is perfectly fine.

This will be summarized in:

```js
--- a/lib/Parser.js
+++ b/lib/Parser.js
@@ -301,7 +301,7 @@ function parseHeader(message, data, headerStart, headerEnd)
       break;
     default:
       // Do not parse this header.
-      message.setHeader(headerName, headerValue);
+      message.addHeader(headerName, headerValue);
       parsed = 0;
   }
```

jmillan:
Good catch @rprinz08,

The solution has been applied for all unknown header fields.

PR #531, opened 2018-09-17
Add missing payload on 'RTCSession:getusermediafailed' event

jonastelzio:
Added missing error parameter on one of the two getusermediafailed-event emits in RTCSession.js


ibc:
Thanks.

ibc:
available in jssip 3.2.12

PR #541, opened 2018-10-11
Consider using pre-loaded adapter.js if present

davies147:
Submitted as requested by @iclems in issue #539


ibc:
Thanks, but instead of this we are just gonna remove adapter.js from JsSIP. The application developer can choose whether using adding it into his application or not.

Issue #538, opened 2018-10-09
Need to emit an event for "out of process" Sip Notify messages


kwleland:
A SIP Message being sent to the UA from the far end server is currently ignored.
It, instead, should cause an event (perhaps newMessage) where the contents of the Notify message
can be retrieved.

This will allow softphones using JsSIP to properly handle
1.) Message Waiting (voicemail) Notifys
2.) BLF (Busy Lamp Field) Notifys

jmillan:
We could create such an event 'newSipEvent' for unsolicited notifications (NOTIFY's that come without subscription).

kwleland:
Beautiful.

spike008t:
I provide a PR to handle NOTIFY event (I do need it).

PR #548, opened 2018-11-30
feat: handle unsubscribe notify event

spike008t:
Here a PR to manage the NOTIFY event.

At the moment is ignored, but Asterisk send a NOTIFY `message-summary` after registered.

The UA emit the new event `sipEvent` with the event and the body to keep it simple.

Should also solve #538

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -666,13 +666,8 @@ module.exports = class UA extends EventEmitter
            */
           break;
         case JsSIP_C.NOTIFY:
-          /**
-           * receive new sip event
-           */
-          this.emit('sipEvent', {
-            event: request.event.event,
-            body: request.body
-          });
+          // receive new sip event
+          this.emit('sipEvent', request);
           request.reply(200);
           break;
         default:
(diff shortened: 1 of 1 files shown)


ibc:
LGTM. @jmillan ?

jmillan:
Thanks!

I'll be some days on holiday. Let me check it once I'm back,

lib/UA.js
```diff
@@ -665,6 +665,16 @@ module.exports = class UA extends EventEmitter
            * and without To tag.
            */
           break;
+        case JsSIP_C.NOTIFY:
+          /**
+           * receive new sip event
+           */
+          this.emit('sipEvent', {
+            event: request.event.event,
```
jmillan: `request.event` provides much also the parameters which are also useful. 

Please change `request.event.event` by `request.event`

lib/UA.js
```diff
@@ -665,6 +665,16 @@ module.exports = class UA extends EventEmitter
...
+            body: request.body
```
jmillan: Providing the full request instead of just the body would be desirable. The user can get the body as well as inspect any SIP header, etc.

jmillan:
Added into master with a slight change. Thanks @spike008t!

PR #550, opened 2018-12-04
fix: works with safari 12.0.1

spike008t:
Hey!

Another PR to support Safari 12 (macOS 10.13.x and iOS 12.x).
Safari do not support anymore `RTCPeerConnection::addStream` and must use `RTCPeerConnection::addTrack`.


lib/RTCSession.js
```diff
@@ -676,7 +676,17 @@ module.exports = class RTCSession extends EventEmitter
         this._localMediaStream = stream;
         if (stream)
         {
-          this._connection.addStream(stream);
+          if (this._connecting.addTrack)
```
ibc: Have you tested this? I don't think `this._connecting.addTrack` is what you want...

lib/RTCSession.js
```diff
@@ -676,7 +676,17 @@ module.exports = class RTCSession extends EventEmitter
...
+          {
+            stream.getTracks().forEach(track => {
+              this._connection.addTrack(track, stream);
+            });
+          }
+
+          if (this._connection.addStream)
+          {
+            this._connection.addStream(stream);
+          }
```
ibc: Why should it call `pc.addStream()` after having called `pc.addTrack()` before?

ibc:
In addition to above referenced issues, what about calls to `this._connection.getLocalStreams()` in RTCSession.js? That's also old stream-based API not supported in Safari.

lib/RTCSession.js
spike008t: arf right! I miss my copy/paste

spike008t: fixed, sorry

ibc:
Note that I did another comment in this conversation.

In addition: is this PR completely tested in Chrome, Firefox and Safari?

lib/RTCSession.js
spike008t: For people having old implementation when listening event `addStream`, they do not have to change the code.

ibc: `pc.onaddstream` event is not fired when calling `pc.addStream()` but just when calling `pc.setRemoteDescription()` with a remote SDP offer/answer that produces remote streams and tracks.

spike008t:
@ibc `RTCPeerConnection::getLocalStreams()` is not supported anymore on safari. 
capture d ecran 2018-12-04 a 22 56 50

That works under safari 12.0.1 on MacOS, Chrome 70.0.3538.110, i'll give a try for firefox.

lib/RTCSession.js
spike008t: Or should I check if `pc.addTrack()` exists, in case of that is not availble, fallback on `pc.addStream()` ?

ibc:
> @ibc `RTCPeerConnection::getLocalStreams()` is not supported anymore on safari.

Right, but RTCSession.js uses it (https://github.com/versatica/JsSIP/blob/master/lib/RTCSession.js#L3318) and this PR does not change it :)

lib/RTCSession.js
ibc: Probably, yes.

spike008t:
> > @ibc `RTCPeerConnection::getLocalStreams()` is not supported anymore on safari.
> 
> Right, but RTCSession.js uses it (https://github.com/versatica/JsSIP/blob/master/lib/RTCSession.js#L3318) and this PR does not change it :)

Oh! i didn't see it... I haven't test mute on my app... I'll give a try

spike008t:
@ibc Ah okay I didn't understand that (must be a quite late at the moment). I'm gonna clean up the code and remove old code with streams

ibc:
Hi. I prefer that @jmillan checks this PR. He is out for two weeks so let's wait a bit, please.

spike008t:
@ibc no problem for checking the PR, in fact I do need to solve my app this week.

davies147:
I though that the whole idea of `webrtc-adapter` was to resolve these compatibility issues so that individual applications such as JsSIP does not have to worry about it.

I understand that the webrtc-adapter dependency was removed from JsSIP's internals, but I believe that was to free up the app developer to use a version of their choice so JsSIP did not have to constantly keep updating.

Or am I missing something? Perhaps I should also be assuming that webrtc-adapter will ensure that new webrtc calls will be polyfilled on older browsers?

If it helps, I just tested un-patched JsSIP on Safari 12.0.1 with adapter.js 6.4.8 and it worked fine.

ibc:
We removed `webrtc-adapter` because, due a bug in the `webrtc-adapter` NPM package` it was not being used at all by JsSIP (`require('webrtc-adapter')` did absolutely nothing).

Even without adapter, JsSIP WebRTC API usage must be modernized. We cannot use the super old stream-based API in 2019.

ibc:
Oh @davies147, super clear now, and agreed :)

nicketson:
I have created a similar PR request here: 

https://github.com/versatica/JsSIP/pull/553

We've been using this fix for some time without issues but haven't had the spare time to create a PR until now.

jmillan:
Corresponding changes have already been applied by #553 

Thanks a lot!

PR #552, opened 2018-12-05
Fixes InviteClientTransaction route_set bugs

nicketson:
ACK and CANCEL requests in an InviteClientTransaction are not
including all of the  Route headers. According to the RFC this is
incorrect. See below:

https://tools.ietf.org/html/rfc3261#section-17.1.1.3

If the INVITE request whose response is being acknowledged had Route
header fields, those header fields MUST appear in the ACK.  This is to
ensure that the ACK can be routed properly through any downstream
stateless proxies.

https://tools.ietf.org/html/rfc3261#section-9.1

If the request being cancelled contains a Route header field, the
CANCEL request MUST include that Route header field's values.

### DIFF
--- a/lib/Transactions.js
+++ b/lib/Transactions.js
@@ -236,7 +236,7 @@ class InviteClientTransaction extends EventEmitter
   {
     const ack = new SIPMessage.OutgoingRequest(JsSIP_C.ACK, this.request.ruri,
       this.ua, {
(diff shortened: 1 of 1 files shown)


ibc:
LGTM but let's please wait 2-3 weeks for @jmillan to check it.

jmillan:
pushed to master. Thanks @nicketson!

PR #553, opened 2018-12-05
Fix: switches to tracks from deprecated stream API

nicketson:
addStream/getLocalStreams usage has been completely removed. The rationale for this is that if someone needs to support an older browser, they can use webrtc-adapter because it provides the necessary shims.

We have been using this fix without any issues for some time.


ibc:
LGTM but let's please wait 2-3 weeks for @jmillan to check it.

jmillan:
Pushed into master. Thanks @nicketson!

PR #554, opened 2018-12-19
fix: add some check on recovery_options interval

spike008t:
Hello,

Here a PR to check the following options : 
- recovery_option.min_interval
- recovery_option.max_interval

That solve an issue I did encountered. I did have a while loop without any interval when retrying to connect.

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -45,6 +45,15 @@ module.exports = class Transport
     this.sockets = [];
 
     this.recovery_options = recovery_options;
+    if (!recovery_options.min_interval || recovery_options.min_interval < 0)
+    {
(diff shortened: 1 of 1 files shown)


jmillan:
This is a bug within your code, which was providing an incorrect `recovery_options` to JsSIP

ibc:
> This is a bug within your code, which was providing an incorrect `recovery_options` to JsSIP

I agree. We cannot sanitize every wrong input the app may attempt to use.

jmillan:
Anyways is good to do some basic checks. Let me check it.

jmillan:
Yeah, well. Lets close it.

PR #559, opened 2019-01-04
Add back support for deprecated stream method

johnryan:
- Add support for deprecated method in older browsers (older versions of chrome/ electron don’t support the addTrack function)
- DRYer code

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -485,6 +485,24 @@ module.exports = class RTCSession extends EventEmitter
     this._progress('local', null);
   }
 
+   /**
+   * Add stream to connection.
+   */
+  _addStream(stream) {
+    if (stream)
+    {
+      if(this._connection.addTrack) {
+        stream.getTracks().forEach((track) =>
+        {
+          this._connection.addTrack(track, stream);
+        });
+      } else {
+        // Support deprecated method in older browsers
+        this._connection.addStream(stream);	
+      }
+    }
+  }
+
   /**
    * Answer the call.
    */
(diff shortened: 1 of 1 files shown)


ibc:
I'm not against this and, in fact, the code looks fine. However, shouldn't this be handled by `webrtc-adapter.js`?

ibc:
I don't mean that JsSIP should neither include `webrtc-adapter` as dependency nor throw an error recommending `webrtc-adapter`. We removed `webrtc-adapter` because it's not needed. And, if needed, the app can include it before loading JsSIP (so JsSIP would transparently use the `RTCPeerConnection` class exposed by adapter.

So the thing is: can you please include adapter in your Electron app and make sure it overrides `window.RTCPeerConnection` before loading JsSIP and then test it?

johnryan:
@ibc So I tested and after importing `webrtc-adapter` in my project it resolves the issue. So as I understand it there are three options:

1) Merge this PR so that `webrtc-adapter` isn't necessary in older chromium versions (at least for the addTrack portion, it seems to work fine without `webrtc-adapter` otherwise)
2) Close this PR and just rely on the developer to patch the issue
3) Provide some sort of error indicator if `!this._connection.addTrack` so that the developer knows what's going wrong (which would be helpful for new users) i.e.:

```
_addStream(stream) {
    if (!this._connection.addTrack) 
    {
         console.error("Unable to add track, please use webrtc-adapter.js");
    } else
    {...}
}
```

Let me know what you think

ibc:
I tend to agree with @rshpount here. The less we hack within JsSIP the less we have to maintain.

johnryan:
@ibc @rshpount I get where you're coming from. I think maybe updating the README / Docs (or even better some sort of console warning?) to advise using `webrtc-adapter` would be helpful for someone who comes across more esoteric bugs like this. As it stands today JsSip will just not work in certain scenarios without `webrtc-adapter` and it's a bit ambiguous as to what actions should be taken to get it to work.

ibc:
We'll keep this PR open to remember adding such a consideration into the README / documentation.

jmillan:
Documentation has been just updated:

https://jssip.net/documentation/misc/interoperability/

PR #563, opened 2019-01-30
use sdpSemantics: 'plan-b'

Krivega:
to support Chrome 72
fix for sdpSemantics: 'unified-plan' in process

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1787,7 +1787,16 @@ module.exports = class RTCSession extends EventEmitter
 
   _createRTCConnection(pcConfig, rtcConstraints)
   {
-    this._connection = new RTCPeerConnection(pcConfig, rtcConstraints);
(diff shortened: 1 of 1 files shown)


Krivega:
because in version 72 the default unified-plan is used. and in the current implementation of the library this causes an erro, because it just uses plan b, although not explicitly.r. it's not about personal preference.

Krivega:
see https://webrtc.org/web-apis/chrome/unified-plan/

rshpount:
@Krivega Can you describe what error this causes?

There was an issue with M72 beta, but as far as I know this was fixed.

Krivega:
@rshpount ok, i will reproduce the error and attach the link

ibc:
> because in version 72 the default unified-plan is used. and in the current implementation of the library this causes an error, because it just uses plan b, although not explicitly. it's not about personal preference.

JsSIP does not assume plan-b or unified-plan.

Issue #566, opened 2019-02-22
Semi-Attended Transfer on ring


michelepra:
I try to manage this situation

1. A call B
2. B answer  (call1)
3. B hold A
3. B dial C (call2)
3. B on RTCSession progress event (180 ringing) refer call2 to call1

jssip not send refer and reporting error on this function

```javascript
  sendRequest(method, options)
  {
    this.debug('sendRequest()');

    return this._dialog.sendRequest(method, options);
  }
```
because this._dialog === undefined.

jmillan:
[Semi-Attended Transfer](https://tools.ietf.org/html/rfc5589#section-7.6) is a **PBX** feature intended to transfer a call even when the Transferor hangs up before the second party answers the call, but the `REFER` SIP message is sent once the dialog is confirmed and not in an early state.

In you patch you are creating a confirmed dialog upon **180** and **183** responses just for your convenience. Apart from being incorrect that would lead into issues when receiving such responses from different endpoints due to parallel forking.

Please use the blind or the attended transfer features which are the ones available at endpoint level.

PR #567, opened 2019-02-22
Change RTCSession to manage Semi-Attended Transfer

michelepra:
issue #566

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1320,7 +1320,9 @@ module.exports = class RTCSession extends EventEmitter
 
     const originalTarget = target;
 
-    if (this._status !== C.STATUS_WAITING_FOR_ACK && this._status !== C.STATUS_CONFIRMED)
+    if (this._status !== C.STATUS_WAITING_FOR_ACK 
+      && this._status !== C.STATUS_CONFIRMED 
+      && this._status !== C.STATUS_1XX_RECEIVED)
     {
       return false;
     }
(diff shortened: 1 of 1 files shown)


lib/RTCSession.js
```diff
@@ -1320,7 +1320,9 @@ module.exports = class RTCSession extends EventEmitter
 
     const originalTarget = target;
 
-    if (this._status !== C.STATUS_WAITING_FOR_ACK && this._status !== C.STATUS_CONFIRMED)
+    if (this._status !== C.STATUS_WAITING_FOR_ACK 
+      && this._status !== C.STATUS_CONFIRMED 
+      && this._status !== C.STATUS_1XX_RECEIVED)
```
ibc: I do not understand this. Why shouldn't the UA be able to refer the remote after receiving a 180 or 183?

michelepra: because jssip allow refer only after receiving 200 OK, if status was STATUS_1XX_RECEIVED ( e.g. 180 Ringing) the next piece of code was `return false;` and this block all refer procedure.

I think this was made because current code does not create RTCSession._dialog when receiving 180

ibc: Let's please wait for @jmillan to come back to activity in 2-3 weeks, he is the mind in this piece of code.

jmillan: I've commented here: https://github.com/versatica/JsSIP/issues/566

jmillan:
Closing as per comments in #566

PR #573, opened 2019-04-10
Fix NameAddrHeader display_name handling

nicketson:
When dealing with NameAddress display names which contained double quotes, such as "John \"The Duke\" Wayne", we have identified several parser issues which prevented proper handling of such display names.

Specifically we identified the following issues:

a. Issues in quoted_string_clean rule:
1. The substring operation (`input.substring(pos-1, offset+1);`) to remove outer quotes returns incorrect results for strings that begin with SWS
2. Strings are not completely unescaped. The outer DQUOTES are removed but quoted_pair sequences are not unescaped.

b. Issues in display_name rule:
1. Quoted display_name is not properly unescaped since quoted_string rule is used.
2. display_name matching token `( LWS token )*` form is not converted to string

c. Issues in NameAddrHeader class: toString() method did not escape display_name

d. UA Config: option validation rejected valid display_name, such as "John \"The Duke\" Wayne".

e. OutgoingRequest: SIPMessage class was creating To and From headers via simple string manipulation and display_name was not properly escaped.

In general handling of quoted strings in display name was inconsistent with double quotes being escaped or passed through depending on the code path.

This pull request addresses all of these issues.

Please note:

1. Other portions of the parser that use the quoted_string rule are also broken because they are not escaped or unescaped, e.g. gen_param. These are not addressed by this pull request.
2. The logic that accepts the integer 0 as a display_name was preserved however it is unclear what purpose/use case it serves.

### DIFF
--- a/lib/NameAddrHeader.js
+++ b/lib/NameAddrHeader.js
@@ -108,9 +108,16 @@ module.exports = class NameAddrHeader
       JSON.parse(JSON.stringify(this._parameters)));
   }
 
+  _quote(str)
+  {
+    return str
+      .replace(/\\/g, '\\\\')
+      .replace(/"/g, '\\"');
+  }
+
   toString()
   {
-    let body = this._display_name ? `"${this._display_name}" ` : '';
+    let body = this._display_name ? `"${this._quote(this._display_name)}" ` : '';
 
     body += `<${this._uri.toString()}>`;
 
--- a/lib/NameAddrHeader.js
+++ b/lib/NameAddrHeader.js
@@ -32,7 +32,7 @@ module.exports = class NameAddrHeader
     // Initialize parameters.
     this._uri = uri;
     this._parameters = {};
-    this._display_name = display_name;
+    this.display_name = display_name;
 
     for (const param in parameters)
     {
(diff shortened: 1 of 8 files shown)


lib/NameAddrHeader.js
```diff
@@ -32,7 +32,7 @@ module.exports = class NameAddrHeader
     // Initialize parameters.
     this._uri = uri;
     this._parameters = {};
-    this._display_name = display_name;
+    this.display_name = display_name;
```
jmillan: No need to remove the underscore. The attribute has its own getter and setter in the class.

jmillan:
I'm fine with it, very nice @nicketson

lib/NameAddrHeader.js
nicketson: I removed the underscore so that `this._display_name` would get assigned via the display_name setter function and `display_name`s of integer 0 would get converted to `"0"`.

This allowed for the removal of duplicated logic handling integer 0 in the toString() method.

jmillan:
New JsSIP release adds this PR.

Thanks @nicketson for the good work!

PR #574, opened 2019-04-17
External timeout to stop ICE candidates gathering

PSyton:
When client has many TURN/STUN servers ICE gathering state should take a
very long time. We should limit this by external timeout for speedup
connection time.

By default this timeout disabled.


ibc:
There is a "candidate" event in RTCSession for that. Just use it.

PR #581, opened 2019-07-01
Export all internal modules in the main JsSIP object

joshfarrant:
A subset of JsSIP's internal modules are already exported on the main JsSIP object, this PR simply exports all of them so that they can be used by clients, if required.

For example, I've been working on adding subscriptions to a client that extends JsSIP, and I currently have to require a handful of files from within JsSIP by referencing them directly.

```js
const DigestAuthentication = require('jssip/lib/DigestAuthentication');
const RequestSender = require('jssip/lib/RequestSender');
const SIPMessage = require('jssip/lib/SIPMessage');
const Transactions = require('jssip/lib/Transactions');
const Transport = require('jssip/lib/Transport');
```

This pull request adds those (and all other) modules to the main JsSIP object so they can instead by used in the following way.

```js
const {
  DigestAuthentication,
  RequestSender,
  SIPMessage,
  Transactions,
  Transport,
} = JsSIP;
```

This shouldn't affect existing functionality, it's simply an enhancement.


jmillan:
Hi @joshfarrant,

Do you think it would be worth it also adding the subscriptions module into JsSIP?

joshfarrant:
Hey @jmillan 👋 

I'm more than happy to share my work on subscriptions, however the code is a little hacky in some places and I don't know whether it's quite PR-ready at the moment. You can find the code [here](https://github.com/Nimvelo/node-client/blob/6b938e12ee7047f52645bcd2a4a68dd857a8c98a/lib/api/webRTC.js).

It relies on JsSIP being passed-in by the client, but ignore that for now. It would work in exactly the same way if JsSIP were required by this module directly.

It would be great for me (and others, I'm sure) to have this built-in to JsSIP. I'd be interested to hear how much additional work you think would need to be done on this for it to be at the point you'd be happy to merge it.

jmillan:
Hi @joshfarrant,

As far as I see you implemented the [dialog](https://tools.ietf.org/html/rfc4235) event package, right?

I don't really know the public API since there is no inline documentation nor the method naming follows the convention used in JsSIP (ie: '_' prefix for private methods).

Could you briefly explain the API of the module?

joshfarrant:
Yes, that's correct.

If you did think it was worth merging then I'd be happy to reformat it to match JsSIP conventions, and include a bit of documentation.

As far as the API goes, it's pretty minimal. I've created a `subscribeToUser` instance method on the UA which accepts a single parameter; the username of the user you'd like to subscribe to. You can call this multiple times to subscribe to multiple users. When the subscription state of any user you've subscribed to changes then a `userStateChanged` event is emitted on the JsSIP UA, which includes both the username of the user whose state has changed, as well as a normalized string describing their new state.

```js
const coolPhone = new JsSIP.UA(configuration);

coolPhone.on('userStateChanged', (username, newState) => {
  // newState can be one of: ['AVAILABLE', 'RINGING', 'BUSY']
  console.log(`User ${username} now has state ${newState}`);
});

coolPhone.on('connected', () => {
  coolPhone.subscribeToUser('012345');
});

coolPhone.start();
```

Hopefully that makes sense - if you have any more questions let me know :+1:

jmillan:
@joshfarrant thanks for your description and sorry for the delay on my response.

I think that conceptually it could be added into JsSIP. Probably the code would need some rework.

Could you please tell me which parts of the [spec](https://tools.ietf.org/html/rfc4235) the implementation misses? Or how compliant it is?

The way I would focus this is making a module (file) that is imported by UA.js. Such file could be named `DialogSubscriber.js`and reside in a new `events` folder: `lib/events/DialogSubscriber.js`. Look [here](https://github.com/versatica/JsSIP/tree/master/lib/RTCSession) as example.

Such a file would implement a class such as `DialogSubscriber` and export the methods:
- `subscribe`
-  `receiveNotify`

UA.js would:
-  import `events/DialogSubscriber` and have an instance of `DialogSubscriber`.
-  export the method `subscribeTo[event]`, in this case: `subscribeToDialog`.
-  export the event `on[event]Notify`, in this case: `onDialogNotify`.

As said, it does conceptually work but the code would require some rework, also considering that ideally there should be a `Subscriber` class from which `DialogSubscriber` and other future event subscribers would inherit.

If you like the idea fell free to make us any questions you may have and we'll try to give you a hand.

jmillan:
closed due to inactivity

joshfarrant:
@jmillan Coming back to this, I've not had any chance to add in the subscriptions module as of yet, but could the original PR be merged in for now?

ibc:
I don't like the idea of this PR, for two reasons:

1. If we publicly expose all those internal classes, then we'll have to keep their internal API and design forever, without breaking changes (since now they are publicly exposed).
2. Importing specific module paths is perfectly fine nowadays. This is:

```js
import * as SIPMessage from 'jssip/lib/SIPMessage';
// or
const SIPMessage = require('jssip/lib/SIPMessage');
```

This is perfectly fine and common nowadays. You need to access some internals of a module to develop something with them, so you just load those internals. Absolutely no need for having all those internal classes exposed into a single `JsSIP` object (which BTW is an anti-pattern in ES6 `import` / `export`).

jmillan:
I would really prefer offering APIs to do things rather than exporting all the internal modules.

This would came with the side effect of having to deal with people making use of this internal code and reporting issues totally related to their code.

Feel free to make a PR with a higher level functionality that we can test and document.

Also agreed with Iñaki's comment.

ibc:
> This is perfectly fine and common nowadays. You need to access some internals of a module to develop something with them, so you just load those internals.

BTW we use this approach in other libraries and code:

https://github.com/versatica/mediasoup-client-aiortc/blob/v3/src/Worker.ts#L4

joshfarrant:
Thanks for your quick responses. @ibc raises some fair concerns which are valid when you're importing JsSIP directly. If, however, you rely on a version of JsSIP to be injected into your client then the suggested solution won't work.

Just to add some background - the reason for wanting these to be exposed through the JsSIP object, rather than importing the modules directly from JsSIP, is that we're currently in a situation where a version of JsSIP needs to be injected into our client. This client is being used on both the web, and in React Native. JsSIP doesn't seem to work out of the box with RN, so we're using a forked version of JsSIP which adds RN support. 

Because our client library needs to work with both the forked (RN-compatible) version of JsSIP, as well as the main published version then we expect the implementer to pass in their preferred version of JsSIP.

A set of APIs to solve this issue would be preferred from my point of view too. For a bit of context, this is where our client implements JsSIP.

https://github.com/sipcentric/node-pbx-client/blob/master/lib/api/webRTC.js

The 5 methods/classes that are not exposed through the main JsSIP import are:

```js
Transactions.checkTransaction();
Transactions.nonInviteServerTransaction();
SIPMessage.OutgoingRequest();
new DigestAuthentication();
new RequestSender();
```

I will note that we're currently importing JsSIP directly into the file linked above, rather than expecting it to be passed in by the implementer.

ibc:
We cannot expose internals just because JsSIP does not work out of the box in RN. If there are issues in RN let's report them, contribute with PR and let's fix them.

Again, exposing the internals make us maintain those classes as they are ad eternum without any chance to refactor them once they are public API. We want to do less, not more.

joshfarrant:
I completely agree with the sentiment that addressing the core issue (adding RN support) is far preferable to adding and maintaining workarounds. I appreciate that exposing these modules could limit the amount that their APIs could change in the immediate future, but I do think _ad eternum_ is a little hyperbolic, as this is why major versions in semver exist.

Thank you both for your thoughts, I'm happy for this PR to remain closed for the time being and we'll figure out another way around the core problem.

ibc:
We don't want to upgrade major version just because, eventually, we want to refactor some internal classes that were never intended to be public. That's the thing. We don't want to get that compromise and that's the extra work for us that I meant.

BTW may be I miss something, but still don't understand why importing module paths is not a solution for your case and why exporting those modules within the main JsSIP object is a solution.

PR #588, opened 2019-08-17
Fix crash in Chromium 61

miachm:
When i try to answer a call in Chromium 61, an exception is raised by a function undefined.

I used this fix to work around the issue, it's working nicely


jmillan:
We are using tracks instead of steams in JsSIP in favour of more modern browsers. For now, I would recommend you to check the browser version and verify the track API existence before JsSIP usage.

jmillan:
@miachm, please read this https://github.com/versatica/JsSIP/pull/559

PR #601, opened 2019-11-12
Add support for rfc 2833

chukitow:
Currently SIPJS supports RFC 2833 for DTMF (https://github.com/onsip/SIP.js/pull/511)
I think will be nice that  JSIP supports it too 😄 

Based on the comments that I found https://github.com/versatica/JsSIP/pull/496 the `sendDtmf`
could take an extra option to define the DTMF type instead of  UA level.


lib/RTCSession.js
```diff
@@ -1042,13 +1045,21 @@ module.exports = class RTCSession extends EventEmitter
...
-
-        options.eventHandlers = {
-          onFailed : () => { this._tones = null; }
-        };
-        dtmf.send(tone, options);
-        timeout = duration + interToneGap;
+        switch (type)
+        {
+          case RFC2833:
+            this._getDTMFRTPSender().insertDTMF(tone, duration, interToneGap);
```
Krivega: if the method returns an undefined? `this._getDTMFRTPSender()`

lib/RTCSession.js
```diff
@@ -3523,4 +3534,21 @@ module.exports = class RTCSession extends EventEmitter
...
     });
   }
+
+  _getDTMFRTPSender()
+  {
+    const streams = this._connection.getLocalStreams();
+
+    if (streams.length > 0)
+    {
+      const audioTracks = streams[0].getAudioTracks();
```
ibc: Honestly I do not like the idea of assuming that there is just a single audio track. Yup, I know that is common in VoIP and SIP regular calls, but forcing JsSIP to assume that by design is not something that makes me too happy.

@jmillan thoughts?

lib/RTCSession.js
chukitow: Good question, I can handle errors using offensive programing like this:
```
 const sender = this._getDTMFRTPSender();
sender && sender.insertDTMF(tone, duration, interToneGap);
```

lib/RTCSession.js
jmillan: I find it OK sending the DTMFs to the **first audio track found**.

Having said that, we no longer use `getLocalStreams()`. See [this].(https://github.com/versatica/JsSIP/blob/master/lib/RTCSession.js#L3346)

Also `createDTMFSender()` is obsoleted in favour of [RTCDTMFSender](https://www.w3.org/TR/webrtc/#dom-rtcdtmfsender)

chukitow:
I updated the PR based on your comments @ibc @jmillan

lib/RTCSession.js
```diff
@@ -3523,4 +3541,14 @@ module.exports = class RTCSession extends EventEmitter
...
   }
+
+  _getDTMFRTPSender()
+  {
+    const dtmfSender = this._connection.getSenders().find((sender) =>
+    {
+      return sender.track && sender.track.kind === 'audio';
+    });
+
+    return dtmfSender && dtmfSender.dtmf;
```
ibc: Can we please avoid this `&&` usage and instead do:

```js
if (dtmfSender)
{
  return dtmfSender.dtmf;
}
```

?

chukitow: Done @ibc

lib/RTCSession.js
```diff
@@ -3523,4 +3541,17 @@ module.exports = class RTCSession extends EventEmitter
...
   }
+
+  _getDTMFRTPSender()
+  {
+    const dtmfSender = this._connection.getSenders().find((sender) =>
+    {
+      return sender.track && sender.track.kind === 'audio';
+    });
+
+    if (dtmfSender)
```
jmillan: Is must also be checked whether the sender can send DTMFs via [canInsertDTMF](https://www.w3.org/TR/webrtc/#dom-rtcdtmfsender-caninsertdtmf) method

``` javascript
if (dtmfSender && dtmfSender.dtmf && dtmfSender.dtmf.canInsertDTMF)
{
 return dtmfSender.dtmf;
}

File: lib/RTCSession.js [hunk 5]
```diff
@@ -39,6 +39,8 @@ const C = {
  */
 const holdMediaTypes = [ 'audio', 'video' ];
 
+const RFC2833 = 'RFC2833';
```
ibc: Do we really need this const? I think it provides no much value and we can get rid of it.

lib/RTCSession.js
```diff
@@ -1042,13 +1045,28 @@ module.exports = class RTCSession extends EventEmitter
...
       else
       {
-        const dtmf = new RTCSession_DTMF(this);
+        switch (type)
+        {
+          case RFC2833: {
+            const sender = this._getDTMFRTPSender();
+
+            if (sender)
+            {
```
ibc: Can we get rid of those unneeded brackets? BTW: have you tested `gulp lint`?

lib/RTCSession.js
jmillan: @ibc I agree with you. There is no need for this const.

To make it more clear to @chukitow,  do you want to share your suggestion for removing this constant?

acharlop:
@ibc should probably close this one

ibc:
Done in 3.4.5

PR #606, opened 2020-03-11
error

lei40251:
When the display_name input the wrong number, the error is reported here


ibc:
If given, `display_number` must be a string, not a number. So that's an usage issue.

PR #619, opened 2020-04-21
sanity: merged request check enhancement

juha-h:
Do not do merged request check if INVITE transaction is already confirmed or non-INVITE transaction is already completed.


juha-h:
I was not able to test if 
```
const Transactions = require('./Transactions');
```
actually worked.

jmillan:
Not needed AFAIK. Fixed in 3.4.3 version.

PR #620, opened 2020-04-22
sanityCheck function rfc3261_8_2_2_2() merged request: check also via branch

juha-h:
Added via_branch check to rfc3261_8_2_2_2() in order to handle delayed removal of transaction.
This is an alternative to earlier pull request that checked transaction status.

### DIFF
--- a/lib/sanityCheck.js
+++ b/lib/sanityCheck.js
@@ -127,6 +127,7 @@ function rfc3261_8_2_2_2()
   const fromTag = message.from_tag;
   const call_id = message.call_id;
   const cseq = message.cseq;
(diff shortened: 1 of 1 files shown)


jmillan:
@juha-h , I resonded in the mailing list. Current behaviour is correct:

>Perhaps in that case the next INVITE comes too fast and the transaction
is still alive.

This happens because the transaction is not immediately removed but upon a timeout (Timer I in RFC3261). Which is value is 0 in reliable transports and T4 in non reliable transports. Hence this is legit, and in UDP this would happen more often.

I'll enhance it so timers with 0 value are executed right away.

PR #624, opened 2020-05-06
Fix to allow initial INVITE no SDP/offer

mgoodenUK:
Allow jssip to respond to INVITE no SDP with a 'a=sendrecv' offer if the user has included rtcOfferConstraints in the session.answer() method.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -500,6 +500,7 @@ module.exports = class RTCSession extends EventEmitter
     const pcConfig = options.pcConfig || { iceServers: [] };
     const rtcConstraints = options.rtcConstraints || null;
     const rtcAnswerConstraints = options.rtcAnswerConstraints || null;
+    const rtcOfferConstraints = options.rtcOfferConstraints || {};
 
     let tracks;
     let peerHasAudioLine = false;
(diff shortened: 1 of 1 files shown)


lib/RTCSession.js
```diff
@@ -617,13 +617,13 @@ module.exports = class RTCSession extends EventEmitter
     }
 
     // Don't ask for audio if the incoming offer has no audio section.
-    if (!mediaStream && !peerHasAudioLine)
+    if (!mediaStream && !peerHasAudioLine && options.rtcOfferConstraints === undefined)
```
ibc, line 620: Please use ! instead (also below).

mgoodenUK: done.

lib/RTCSession.js
```diff
@@ -617,13 +617,13 @@ module.exports = class RTCSession extends EventEmitter
     }
 
     // Don't ask for audio if the incoming offer has no audio section.
-    if (!mediaStream && !peerHasAudioLine)
+    if (!mediaStream && !peerHasAudioLine && !options.rtcOfferConstraints)
```
jmillan, line 620: We need to check *specifically* if `offerToReceiveAudio` constraint is set to not.

Also consider that `rtcOfferConstraints` is optional and it may not be present. You can create a variable as it is done for `rtcAnswerConstraints` [here](https://github.com/versatica/JsSIP/pull/624/files#diff-f0bb4b4297430ea5a3622325573c2ed9R502)

jmillan:
Nice, I'll merge it and create a new release when https://github.com/versatica/JsSIP/pull/625 is finished.

jmillan:
thanks @mgoodenUK,

The PR has been merged and a is available in a new release :-)

PR #625, opened 2020-05-07
clone mediaConstraints options

acharlop:
otherwise an error is thrown trying to assign to a frozen object and the value is already set.

For example if sending options:

`{ mediaConstraints: { audio: true, video: false }}`

This code block throws an error:

```
// Don't ask for video if the incoming offer has no video section.
if (!mediaStream && !peerHasVideoLine)
{
    mediaConstraints.video = false;
}
```

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -266,11 +266,14 @@ module.exports = class RTCSession extends EventEmitter
     debug('connect()');
 
     const originalTarget = target;
-    const eventHandlers = options.eventHandlers || {};
+    const eventHandlers = Utils.cloneObject(options.eventHandlers);
     const extraHeaders = Utils.cloneArray(options.extraHeaders);
-    const mediaConstraints = options.mediaConstraints || { audio: true, video: true };
+    const mediaConstraints = Utils.cloneObject(options.mediaConstraints, {
+      audio : true,
+      video : false
+    });
     const mediaStream = options.mediaStream || null;
-    const pcConfig = options.pcConfig || { iceServers: [] };
+    const pcConfig = Utils.cloneObject(options.pcConfig, { iceServers: [] });
     const rtcConstraints = options.rtcConstraints || null;
     const rtcOfferConstraints = options.rtcOfferConstraints || null;
 
@@ -497,7 +500,7 @@ module.exports = class RTCSession extends EventEmitter
     const extraHeaders = Utils.cloneArray(options.extraHeaders);
     const mediaConstraints = Utils.cloneObject(options.mediaConstraints);
     const mediaStream = options.mediaStream || null;
-    const pcConfig = options.pcConfig || { iceServers: [] };
+    const pcConfig = Utils.cloneObject(options.pcConfig, { iceServers: [] });
     const rtcConstraints = options.rtcConstraints || null;
     const rtcAnswerConstraints = options.rtcAnswerConstraints || null;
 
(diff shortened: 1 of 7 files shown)


jmillan:
Can you attach the given JsSIP console logs with the error?

acharlop:
```
Uncaught TypeError: Cannot assign to read only property 'video' of object '#<Object>'
    at RTCSession.answer (RTCSession.js:578)
```

ibc:
Please show also your code (just related code), specially how do you create those options and how do you modify them later.

acharlop:
```
// in redux
export const INITIAL_STATE = {
  sessionOptions = {
    mediaConstraints: {
      audio: true,
      video: false,
    },
    pcConfig: {
      iceServers: [
        {
          urls: [],
        },
      ],
    },
  }
}

// in reducer
const { config: { sessionOptions }} = getState()
session.answer(sessionOptions)
```

ibc:
I don't think that code snippet reflects the issue. Please avoid React/Redux syntax and just show the usage of the `{ mediaConstraints: { audio: true, video: false }}` object, how you pass it to JsSIP and how you later try to modify it.

acharlop:
I don't try to modify the object, I create it once and pass it to JsSIP when necessary.

The error being thrown in from the JsSIP lib `RTCSession.js:578` where the following happens.

line 498:
```
const mediaConstraints = options.mediaConstraints = {}
```

578:
```
// Don't ask for video if the incoming offer has no video section.
if (!mediaStream && !peerHasVideoLine)
{
    mediaConstraints.video = false;
}
```

This is where the assigning to a frozen object happens

ibc:
@jmillan are we freezing that object at all in JsSIP?

I don't think so. Maybe the problem is that you are passing a frozen object, right?

acharlop:
correct. My suggestion is that JsSIP shouldn't rely on the passed in object to not be frozen, just clone it and 💪

ibc:
Not against that since it makes sense. However, this PR just covers your exact needs. The very same could happen with other similar objects passed to JsSIP via API. Perhaps could you extend this PR to also cover other cases?

jmillan:
> Not against that since it makes sense. However, this PR just covers your exact needs. The very same could happen with other similar objects passed to JsSIP via API. Perhaps could you extend this PR to also cover other cases?

That would be nice, indeed.

acharlop:
will do 👍

acharlop:
Updated. Added an option for fallback option if `{}` is not the desired default

lib/RTCSession.js
```diff
@@ -266,11 +266,14 @@ module.exports = class RTCSession extends EventEmitter
...
 
     const originalTarget = target;
-    const eventHandlers = options.eventHandlers || {};
+    const eventHandlers = Utils.cloneObject(options.eventHandlers);
     const extraHeaders = Utils.cloneArray(options.extraHeaders);
-    const mediaConstraints = options.mediaConstraints || { audio: true, video: true };
+    const mediaConstraints = Utils.cloneObject(options.mediaConstraints, {
+      audio : true,
+      video : false
+    });
```
ibc, line 274: this is wrong, default value for `video` is true.

acharlop: fixed thanks

ibc:
LGTM, @jmillan ?

lib/SIPMessage.js
```diff
@@ -25,7 +25,7 @@ class OutgoingRequest
       return null;
     }
 
-    params = params || {};
+    params = Utils.cloneObject(params);
```
jmillan, line 28: this one is not needed. `params` are internally generated.

acharlop: updated

jmillan:
thanks @acharlop,

The PR has been merged and a is available in a new release :-)

PR #627, opened 2020-05-08
Typescript typings

phwb:
Hi! I've put together some Typescript typings. Please review. Not all types covered, I will continue work on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "contributors": [
     "Iñaki Baz Castillo <ibc@aliax.net> (https://github.com/ibc)"
   ],
+  "types": "lib/JsSIP.d.ts",
   "main": "lib-es5/JsSIP.js",
   "keywords": [
     "sip",
@@ -26,6 +27,8 @@
     "url": "https://github.com/versatica/JsSIP/issues"
   },
   "dependencies": {
+    "@types/debug": "^4.1.5",
+    "@types/node": "^14.0.6",
     "debug": "^4.1.1",
     "events": "^3.1.0",
     "sdp-transform": "^2.14.0"
--- a/lib/Constants.d.ts
+++ b/lib/Constants.d.ts
@@ -1,0 +1,66 @@
+export const USER_AGENT: string
+export const SIP = 'sip'
+export const SIPS = 'sips'
+
+export enum causes {
+  CONNECTION_ERROR = 'Connection Error',
+  REQUEST_TIMEOUT = 'Request Timeout',
+  SIP_FAILURE_CODE = 'SIP Failure Code',
+  INTERNAL_ERROR = 'Internal Error',
+  BUSY = 'Busy',
+  REJECTED = 'Rejected',
+  REDIRECTED = 'Redirected',
+  UNAVAILABLE = 'Unavailable',
+  NOT_FOUND = 'Not Found',
+  ADDRESS_INCOMPLETE = 'Address Incomplete',
+  INCOMPATIBLE_SDP = 'Incompatible SDP',
+  MISSING_SDP = 'Missing SDP',
+  AUTHENTICATION_ERROR = 'Authentication Error',
+  BYE = 'Terminated',
+  WEBRTC_ERROR = 'WebRTC Error',
+  CANCELED = 'Canceled',
+  NO_ANSWER = 'No Answer',
+  EXPIRES = 'Expires',
+  NO_ACK = 'No ACK',
+  DIALOG_ERROR = 'Dialog Error',
+  USER_DENIED_MEDIA_ACCESS = 'User Denied Media Access',
+  BAD_MEDIA_DESCRIPTION = 'Bad Media Description',
+  RTP_TIMEOUT = 'RTP Timeout',
+}
+
+export const SIP_ERROR_CAUSES: {
+  REDIRECTED: [300, 301, 302, 305, 380],
+  BUSY: [486, 600],
+  REJECTED: [403, 603],
+  NOT_FOUND: [404, 604],
+  UNAVAILABLE: [480, 410, 408, 430],
+  ADDRESS_INCOMPLETE: [484, 424],
+  INCOMPATIBLE_SDP: [488, 606],
+  AUTHENTICATION_ERROR: [401, 407]
+}
+export const ACK = 'ACK'
+export const BYE = 'BYE'
+export const CANCEL = 'CANCEL'
+export const INFO = 'INFO'
+export const INVITE = 'INVITE'
+export const MESSAGE = 'MESSAGE'
+export const NOTIFY = 'NOTIFY'
+export const OPTIONS = 'OPTIONS'
+export const REGISTER = 'REGISTER'
+export const REFER = 'REFER'
+export const UPDATE = 'UPDATE'
+export const SUBSCRIBE = 'SUBSCRIBE'
+
+export enum DTMF_TRANSPORT {
+  INFO = 'INFO',
+  RFC2833 = 'RFC2833',
+}
+
+export const REASON_PHRASE: Record<number, string>
+export const ALLOWED_METHODS = 'INVITE,ACK,CANCEL,BYE,UPDATE,MESSAGE,OPTIONS,REFER,INFO,NOTIFY'
+export const ACCEPTED_BODY_TYPES = 'application/sdp, application/dtmf-relay'
+export const MAX_FORWARDS = 69
+export const SESSION_EXPIRES = 90
+export const MIN_SESSION_EXPIRES = 60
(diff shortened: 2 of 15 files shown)


jmillan:
Looks very nice!

Is all the API covered?

jmillan:
> Is all the API covered?

Oh, you already said it's a work in progress. Nice!

phwb:
I covered only main API

image

jmillan:
What about having each type in the corresponding file and have a index.d.ts exporting them all?

So we would have `RTCPeerConnection.d.ts`, `UA.d.ts` etc.

jmillan:
> I covered only main API

:+1:

looking good so far.

phwb:
> What about having each type in the corresponding file and have a index.d.ts exporting them all?

Honestly, i don't think about it, may be do it in next PR 🙂

jmillan:
> Honestly, i don't think about it, may be do it in next PR 🙂

Are you thinking on completing the API in this PR?

phwb:
> Are you thinking on completing the API in this PR?

Yes. I already use this types in production

jmillan:
> Yes. I already use this types in production

I mean whether you are willing to complete the full public API in this PR. Since you say not all types are covered.

phwb:
Hi, I fully covered documented API and split in to separate files 🙂

.gitignore
```diff
@@ -1,2 +1,3 @@
+.idea
```
jmillan, line 1: Can you please remove this from here and add it in your personal .gitignore instead?

jmillan:
Let us some days to review this. 

Looks great BTW.

.gitignore
phwb: fixed

jmillan:
@ibc can you please check it? It looks good to me so far.

ibc:
While this is a nice work (thanks a lot) I'm afraid this will be hard to maintain. For instance, it misses new API introduced in PR https://github.com/versatica/JsSIP/pull/629. Can you add it please?

I think we should eventually rewrite the whole lib in TypeScript so it would be easier to maintain.

phwb:
> I think we should eventually rewrite the whole lib in TypeScript so it would be easier to maintain.

It will be great. I think, i can help with it.

> LGTM. Just please add the new transportType in sendDTMF() which has been introcuded today.

ок

phwb:
> add the new transportType in sendDTMF()

done

package.json
```diff
@@ -26,6 +27,8 @@
     "url": "https://github.com/versatica/JsSIP/issues"
   },
   "dependencies": {
+    "@types/debug": "^4.1.5",
+    "@types/node": "^13.13.5",
```
ibc, line 31: `@types/debug` must be in `dependencies` because JsSIP exports a `debug` instance. But why is `@types/node` there? It should be in `devDependencies`.

phwb: Because some classes extends from EventEmitter. And if `@types/node` were in `devDependencies`, types for `EventEmitter` wouldn't be installed

ibc: This is true, but note that JsSIP uses the [events](https://github.com/Gozala/events) package, so we should have `@types/events` in `dependencies` rather than `@types/node`, right?

phwb: This package realize Node interface, so it's correct

phwb: I checked it with Typescript

ibc: It may work, but if JsSIP runs in browser we don't need Node types. Can you try by uninstalling it and installing `@types/events` instead?

phwb: I already tried. Create empty project, install `events` and `@types\events`, create test file:
```jsx
import {EventEmitter} from 'events';

export class Test extends EventEmitter {}
```
then compiled it, and receive next declaration:
```jsx
/// <reference types="node" />
import { EventEmitter } from 'events';
export declare class Test extends EventEmitter {
}
```

ibc: Not super expert in TS here. So what? is it ok or not?

phwb: It's ok

phwb: `@types/node` and `@types/events` declaration are the same

ibc: Let's add both them because JsSIP works in browser and Node.

phwb: It doesn't matter, because TS use only `@types/node`. And it's just interface

phwb: If I add `@types/node`, then it will be installed, but not used

phwb: So, what we gona do? Leave `@type/node` or replace to `@types/events`?

phwb: @ibc are you forget about me? 🙂

ibc: We did not, it's just very bad days. Please let's come back to this next week. We don't forget.

phwb: ok, no problem

ibc:
@jmillan can we discuss this tomorrow?

package.json
jmillan: if @types/node includes events definitions then lets use the first one only :+1:

ibc, line 31: Please use `@types/node" 14.x.x, the latest stable version.

phwb: done

jmillan:
Merged into master,

Thanks @phwb :-)

jmillan:
@phwb, we'll announce the TS support in JsSIP and want to reference you if you are OK with it.

Do you have a Twitter account or do you prefer to point to you Github one?

phwb:
> @phwb, we'll announce the TS support in JsSIP and want to reference you if you are OK with it.
> Do you have a Twitter account or do you prefer to point to your Github one?

Yes, it will be nice 🙂 you can use my Github account

jmillan:
There you are: https://twitter.com/jomivi/status/1268085538042544133

Thanks!

arkaev:
Hello.
Decision to add @types/node is totally wrong. It breaks types for projects that don't use it. Please remove it

phwb:
> Decision to add @types/node is totally wrong. It breaks types for projects that don't use it. Please remove it

Hi,
Can you show breaking code or repo?

arkaev:
For example:
  "dependencies": {
    "jssip": "^3.3.11"
  },
  "devDependencies": {
    "@types/events": "^3.0.0"
  }

import { Listener } from 'events';
^
Module '"events"' has no exported member 'Listener'

And many issues in TypeScript github about this situation. Official position of this command - fix dependency in library
https://github.com/microsoft/TypeScript/issues/18588
https://github.com/microsoft/TypeScript/issues/17042

phwb:
> And many issues in TypeScript github about this situation. Official position of this command - fix dependency in library

thanks, i will look at this soon

jmillan:
Adding it as `devDependency` would be the way to go in JsSIP, right?

arkaev:
I can't check but looks like devDependency is good move

phwb:
@arkaev you can install manually `@types/events`, it fix you problem.

arkaev:
@phwb  it's just an example. Problem more complex with many conflicts... And can be solved by typeRoots with manually fixed types. But it's ugly and unmantainable

phwb:
@arkaev i create PR https://github.com/versatica/JsSIP/pull/645 to fix this problem

jmillan:
So basically @types/node is the culprit and should not be used as dependency?

phwb:
> So basically @types/node is the culprit and should not be used as dependency?

Sounds like yes, it should be replaced. 

I did it because typescript compiler did it. 
I created empty project, add `events` and `@types/events` as dependencies, create `.ts` file:
```jsx
import {EventEmitter} from 'events'
``` 
compile `tsc -d` and got the following file:
```jsx
/// <reference types="node" />
import {EventEmitter} from 'events'
```
`/// <reference path="..." />` directive needs `@types/node` as deps

phwb:
Although... If you want to rewrite all code into typescript, then you got same result
```jsx
/// <reference types="node" />
import {EventEmitter} from 'events'
```
and `@types/node` is correct

phwb:
@arkaev I think, that replace `@types/node` to `@types/events`, doesn't solve you problem

arkaev:
devDependencies should solve problem

phwb:
> devDependencies should solve problem

no, because `tsc` write `/// <reference types="node" />` in `.d.ts` file, when project migrate to typescript

arkaev:
Do you use something from node typings? Found regexpp package with @types/node as devDepenency but without <reference> in typings

phwb:
@arkaev my apologies, i mistake. I created an empty project one more time, compile it and tsc doesn't write  `/// <reference types="node" />` to `.d.ts` file. I will reopen PR with fix.

phwb:
https://github.com/versatica/JsSIP/pull/648

PR #628, opened 2020-05-08
refactor: [clone-media-constraints-cleanup] - clone object added in PR 624 - and released in 3.4.4

acharlop:
Just to keep things clean 👍


jmillan:
Added to master, will be available on the next release.

PR #633, opened 2020-05-20
feat: add replaceMediaStream to session

Krivega:


jmillan:
Hi,

I would rather don't renegotiate inside `replaceMediaStream` and let it just replace the media stream.

lib/RTCSession.js
```diff
@@ -1344,6 +1342,38 @@ module.exports = class RTCSession extends EventEmitter
...
 
+  replaceMediaStream(mediaStream, options)
+  {
+    debug('replaceMediaStream()');
+
+    return new Promise((resolve, reject) =>
+    {
+      this._removeMediaStream(this._localMediaStream);
+      this._addMediaStream(mediaStream);
+      this.renegotiate(options, resolve, reject);
```
jmillan, line 1353: call to `renegotiate` should go out of this method.

lib/RTCSession.js
```diff
@@ -1344,6 +1342,38 @@ module.exports = class RTCSession extends EventEmitter
...
+    });
+  }
+
+
+  _addMediaStream(stream)
+  {
+    stream.getTracks().forEach((track) => this._connection.addTrack(track, stream));
```
jmillan, line 1360: Please respect the current style, use braces as in the rest of the code.

lib/RTCSession.js
```diff
@@ -1344,6 +1342,38 @@ module.exports = class RTCSession extends EventEmitter
...
+
+  _removeMediaStream(stream)
+  {
+    this._connection.getSenders().forEach((sender) =>
+    {
+      stream.getTracks().forEach((track) =>
+      {
+        if (sender.track === track)
+        {
+          this._connection.removeTrack(sender);
```
jmillan, line 1371: `break` here.

lib/RTCSession.js
```diff
@@ -1821,7 +1851,16 @@ module.exports = class RTCSession extends EventEmitter
 
   _createRTCConnection(pcConfig, rtcConstraints)
   {
-    this._connection = new RTCPeerConnection(pcConfig, rtcConstraints);
+    this._connection = new RTCPeerConnection(
```
jmillan, line 1854: This change has nothing to do here.

Krivega: also surprised

Krivega:
> Hi,
> 
> I would rather don't renegotiate inside `replaceMediaStream` and let it just replace the media stream.

fail to renegotiate leave?

lib/RTCSession.js
```diff
@@ -1290,7 +1287,7 @@ module.exports = class RTCSession extends EventEmitter
     return true;
   }
 
-  renegotiate(options = {}, done)
+  renegotiate(options = {}, done, fail)
```
jmillan, line 1290: `fail` needs to go away. Failures are handled by terminating the session.

lib/RTCSession.js
Krivega: why do you need to break?

jmillan: There is one track for sender. If you found it then there's no reason to keep iterating.

jmillan: wait, we are inside a `forEach`

Krivega: yes, it needs to be rewritten to `for()`

jmillan:
```js
if (stream.getTracks().includes(sender.track) {
  this._connection.removeTrack(sender);
}
```

jmillan: no need for the nested `forEach`

Krivega: I thought about
```js
    stream.getTracks().forEach((track) => {
      const senderMediaStream = this._connection.getSenders().find((sender) => sender.track === track);

      if (senderMediaStream) {
        this._connection.removeTrack(senderMediaStream);
      }
    });
``` but it's better

Comment (Krivega) [reply lib/RTCSession.js]: or 
```js

    this._connection.getSenders()
    .filter((sender) => stream.getTracks().includes(sender.track))
    .forEach((sender) => { 
      this._connection.removeTrack(sender);
    });
```

ibc:
Honestly I don't like this PR much. We don't need an API to replace the current `MediaStream`. The application can get the `RTCPeerConnection` via `session.connection` and then use the WebRTC API to replace the audio or the video track using (for example):

```js
const rtpSender = session.connection.getSenders()[0];

await rtpSender.replaceTrack(newAudioTrack)`.
```

So thanks for this effort but I'm against adding sugar API to JsSIP (which means lot of maintenance for us).

jmillan:
The negotiation has been removed from the PR. I asked for that. There has been some other cleanup.

I had the same feeling @ibc, but if you look at the changes now it just does that; remove tracks from current stream and add tracks from a new one. It's a single new public method,  syntactic sugar but its so harmless, and replacing the stream is something for which actually we get questions about in the mailing list.

Please review the changes again.

ibc:
> syntactic sugar but its so harmless, and replacing the stream is something for which actually we get questions about in the mailing list.

Why do we assume that we just handle a single `MediaStream`? We may handle an audio track from stream1 and video track from stream2.

ibc:
> replacing the stream is something for which actually we get questions about in the mailing list

Because people do not take a look to the WebRTC API and avoid using `session.connection`. That does not mean that we should provide a replacement for them.

But ok, changes in this PR look much tiny now. However, TypeScript required.

KraftyKraft:
> @KraftyKraft can you also send a PR to https://github.com/versatica/jssip.net with the new API?

Hi! Did you mean to mention @Krivega ? :)

ibc:
Sure, crazy GitHub nick suggestions :)

Krivega:
sure

ibc:
> can you also send a PR to https://github.com/versatica/jssip.net with the new API?

Hi @Krivega, waiting for this in order to accept the PR :)

Krivega:
@ibc with `replaceTrack` switching the track should not require negotiation
but I have doubts about adding a track, to be consistent, you need to check tracks that we did not replace and delete. Or, at the beginning, check the current tracks and compare with those that come and in case of mismatch give an error / warning

ibc:
> @ibc with `replaceTrack` switching the track should not require negotiation
> but I have doubts about adding a track, to be consistent, you need to check tracks that we did not replace and delete. Or, at the beginning, check the current tracks and compare with those that come and in case of mismatch give an error / warning

That is exactly one of the things I just don't want to do within JsSIP internals. If the app adds a new track it should renegotiate. It the app replaces a track, it shouldn't.

jmillan:
Guys,

This was a great attempt to create a higher level public method for a common usage, but it has demonstrated to break the paradigm of the library adding a complex to comprehend and maintain method that would require just few lines of code in de the app.

@Krivega, thanks for the efforts, I hope you understand the decision to close the PR. :+1:

PR #640, opened 2020-07-07
Content-Type is case insensitive

hxl-dy:


jmillan:
Nice,

It would be more reliable if the parser already returned a lowercase value:

[here](https://github.com/versatica/JsSIP/blob/master/lib/Grammar.pegjs#L507)

```
data = input.substring(pos, offset).toLowerCase();
````

After updating Grammer.pegjs file you need to run `gulp grammar`  in order to generate te corresponding new grammar js file.

**Comment by hxl-dy** [comment]
> Nice,
> 
> It would be more reliable if the parser already returned a lowercase value:
> 
> [here](https://github.com/versatica/JsSIP/blob/master/lib/Grammar.pegjs#L507)
> 
> ```
> data = input.substring(pos, offset).toLowerCase();
> ```
> 
> After updating Grammer.pegjs file you need to run `gulp grammar` in order to generate te corresponding new grammar js file.

done

**Comment by jmillan** [comment]
The only change needed really is:

`data = input.substring(pos, offset).toLowerCase();`  [here](https://github.com/versatica/JsSIP/blob/master/lib/Grammar.pegjs#L507)

**Comment by hxl-dy** [comment]
> The only change needed really is:
> 
> `data = input.substring(pos, offset).toLowerCase();` [here](https://github.com/versatica/JsSIP/blob/master/lib/Grammar.pegjs#L507)

only changing this does not work

**Comment by jmillan** [comment]
you're right..

Then sorry.., the initial proposal was OK.

Could you please revert?

**Comment by hxl-dy** [comment]
It's not necessary to revert. The lastest proposal use Grammar+parseHeader to get the parsed header.

**Comment by jmillan** [comment]
> It's not necessary to revert. The lastest proposal use Grammar+parseHeader to get the parsed header.

I know, but it's semantically incorrect `parsing` an already parsed string.

Your initial implementation was OK.

**Comment by jmillan** [comment]
Thanks @hxl-dy!

File: lib/RTCSession.js [hunk 1]
```diff
@@ -390,7 +390,7 @@ module.exports = class RTCSession extends EventEmitter
     debug('init_incoming()');
 
     let expires;
-    const contentType = request.getHeader('Content-Type');
+    const contentType = request.getHeader('Content-Type').toLowerCase();
```
ibc, line 393: Guys, what will happen if the request does not have `Content-Type` header? It will crash.

hxl-dy: `String(request.getHeader('Content-Type')).toLowerCase()` can fix it.

ibc: `String(undefined)` returns `"undefined"` string. Not a valid solution. Existence of the header must be checked before getting its value. Can you do a PR for it, please?

Issue #499, opened 2018-01-31
RE-INVITE sent before ICE gathering is complete


shazzzzzzzz:
**scenario 1-** 

- A calls B. 
     - Audio only. 
     - A supports Bundle and Trickle ICE. 
     - No Bundle or Trickle ICE support on B.
- A turns on camera. 
     - Application internally adds a video track to the existing stream on A and starts re-negotiation.

**Expected behavior** - Since JsSIP does not support trickle ICE, it must trigger a RE-INVITE once it gathers all the ICE candidate for the new video track.
**Observed behavior** - RE-INVITE is sent out before new ICE candidates could be gathered. B is unaware of A's candidates, hence no RTP streaming possible.

From my code analysis(if it helps) - When A triggers re-negotiation, iceGatheringState is already "complete" upon setLocalDescription invocation. Due to this create offer's success callback is invoked before ICE gathering is actually completed. 

This could impact another scenario as well as below
**Scenario 2-**

- B calls A. Audio 'sendrecv' and video 'inactive'
- B turns on camera. Replaces existing stream with a new stream containing video tracks and triggers re-negotiation with new ICE candidates
- A responds back with 200 OK, however I still see ICE gathering being completed after 200 OK is sent out.

**Expected behavior** - 
200 OK must be sent after ICE gathering is complete
                                       OR 
Should 'A' even gather ICE candidates? since it could use the same candidates from initial call setup. With scenario 2(B calls A. Audio 'sendrecv' and video 'inactive') and A turning on camera, I still see new ICE candidates being gathered for RE-INVITE to be sent out. I am not sure on the expected behavior here. 

Attached are the browser logs for both scenarios on A. 

[scenario_1.txt](https://github.com/versatica/JsSIP/files/1682353/scenario_1.txt)
[scenario_2.txt](https://github.com/versatica/JsSIP/files/1682354/scenario_2.txt)

jmillan:
Hi,

May you register a callback to the `RTCSession::connection` ` negotiationneeded` event once you add the video track and actually renegotiate from such callback? Looks like the peerconnection is not totally ready yet when re-inviting. Ie:


```javascript
function renegotiate()
{
  session.renegotiate();
  session.connection.removeEventListener('negotiationneeded', renegotiate);
}

// Add video track here...

session.connection.addEventListener('negotiationneeded', renegotiate);
```

Thanks

shazzzzzzzz:
Hi,

I tried the solution you suggested and here are the results.

[...]

PR #641, opened 2020-07-07
Force gather candidates when iceRestart is true

jose-lopes:
Hello,

Before I describe the pull request, I want to thank you for the good work on the project JsSIP.

I had some problems with network switching (eg: between wireless networks), when I try to use the next constraint: iceRestart true .
I notice the SDP was recreated without ice candidates, because the ICE Candidates Gathering was not executed, related to iceGatheringState === "complete"
This issue is similar to last scenario described at https://github.com/versatica/JsSIP/issues/499.
So I created an exception when there is constraint iceRestart: true, to continue to execution of ICE Candidates Gathering.
This commit solve the issue.

Thanks for the attention and I will appreciate some feedback/guidance .

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1903,7 +1903,7 @@ module.exports = class RTCSession extends EventEmitter
       .then(() =>
       {
         // Resolve right away if 'pc.iceGatheringState' is 'complete'.
-        if (connection.iceGatheringState === 'complete')
+        if (connection.iceGatheringState === 'complete' && (!constraints || !constraints.iceRestart))
         {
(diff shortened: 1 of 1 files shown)


jmillan:
Looks good.

Just make sure to not access a member of `constraints` if it's undefined:

`if (connection.iceGatheringState === 'complete' && (!constraints || !constraints.iceRestart)`

jose-lopes:
@jmillan Thanks for the feedback, I have done the change.

jmillan:
Thanks @jose-lopes,

Let's wait a bit for #640 and then create a new release.

PR #647, opened 2020-07-09
UA: fix UAConfiguration.sockets TypeScript typings

lei40251:
UA: fix UAConfiguration.sockets TypeScript typings

### DIFF
--- a/lib/UA.d.ts
+++ b/lib/UA.d.ts
@@ -18,14 +18,9 @@ export interface CallOptions extends AnswerOptions {
   anonymous?: boolean;
 }
 
-export interface Sockets {
-  socket: Socket;
-  weight: number
-}
-
 export interface UAConfiguration {
   // mandatory parameters
-  sockets: Socket | Socket[] | Sockets;
+  sockets: Socket | Socket[] | WeightedSocket ;
   uri: string;
   // optional parameters
   authorization_user?: string;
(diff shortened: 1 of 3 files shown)


lib/UA.d.ts
```diff
@@ -18,9 +18,14 @@ export interface CallOptions extends AnswerOptions {
   anonymous?: boolean;
 }
 
+export interface Sockets {
```
jmillan, line 21: `Sockets` semantic implies multiple `Socket`s. And this is a different type. You can call it `WeightedSocket`.

jmillan: Also, please move it to WebSocketInterface file as `Socket` definition resides there.

jmillan: And must also be exported in JsSIP.d.ts

lib/UA.d.ts
```diff
@@ -18,14 +18,9 @@ export interface CallOptions extends AnswerOptions {
...
 
-export interface Sockets {
-  socket: Socket;
-  weight: number
-}
-
 export interface UAConfiguration {
   // mandatory parameters
-  sockets: Socket | Socket[] | Sockets;
+  sockets: Socket | Socket[] | WeightedSocket ;
```
jmillan, line 23:
```suggestion
  sockets: Socket | Socket[] | WeightedSocket[] ;
```

jmillan:
thanks @lei40251

PR #652, opened 2020-07-15
UA: Allow to change authorization_user

jose-lopes:
Hello,

I am using auth ephemeral to integrate with module auth_ephemeral from Kamailio.(https://kamailio.org/docs/modules/stable/modules/auth_ephemeral.html)

The credentials are only valid for a short period of time, so I need to update the authorization_user and password.
This pull request allow to to update the authorization_user.

Thanks for the attention and I will appreciate some feedback/guidance.

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -355,6 +355,9 @@ module.exports = class UA extends EventEmitter
   {
     switch (parameter)
     {
+      case 'authorization_user':
+        return this._configuration.authorization_user;
+
       case 'realm':
         return this._configuration.realm;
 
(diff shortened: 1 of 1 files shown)


lib/UA.js
```diff
@@ -887,7 +895,7 @@ module.exports = class UA extends EventEmitter
 
     // Seal the configuration.
     const writable_parameters = [
-      'password', 'realm', 'ha1', 'display_name', 'register'
+      'password', 'realm', 'ha1', 'display_name', 'register', 'authorization_user'
```
jmillan, line 898: Can you pleas move 'authorization_user' to the first position for consistency?

jose-lopes: Sure, I will do it

jmillan:
Merge, thanks @jose-lopes

PR #653, opened 2020-07-23
cseq.value->cseq in Dialog/RequestSender.js

mattdimeo:
Fixes an exception in Dialog/RequestSender.js when in a 491-state transaction.


jmillan:
Merged, thanks @mattdimeo

PR #655, opened 2020-07-29
emit data when parse message fails

lazedo:

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -1013,6 +1013,7 @@ function onTransportData(data)
 
   if (! message)
   {
+    this.emit('data', data.message);
     return;
   }
 


jmillan:
Please, explain the rationale of this PR.

jmillan:
wouldn't a log line do the work?

lazedo:
@jmillan Hello, the intent is to handle `non-sip` payloads when using the same socket for sip & json events

jmillan:
@lazedo,

This is why there is a [Socket interface](https://jssip.net/documentation/3.4.x/api/socket/) so you can implement your own socket that handles SIP and no SIP traffic.

If any doubt please ask in the mailing list.

PR #671, opened 2020-11-02
Add a config parameter to set the session timer refresher in the first request sent by JsSip

kkozlik:
ATM, JsSIP support the SIP session timers, and works strictly according to the RFC, which gives preference to the UAS triggering the session refresh. However in environment where the WebRTC client is connected via a client-side only initiated connection, it makes sense to force JsSIP to be the refresher of session timer. All it takes is basically to add the "refresher=uac" parameter to the Session-Expires HF in the initial INVITE

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -25,9 +25,10 @@ exports.settings = {
   use_preloaded_route : false,
 
   // Session parameters.
-  session_timers                : true,
-  session_timers_refresh_method : JsSIP_C.UPDATE,
-  no_answer_timeout             : 60,
+  session_timers                 : true,
+  session_timers_refresh_method  : JsSIP_C.UPDATE,
+  session_timers_force_refresher : false,
+  no_answer_timeout              : 60,
 
   // Registration parameters.
   register         : true,
@@ -230,6 +231,14 @@ const checks = {
         {
           return method;
         }
+      }
+    },
+
+    session_timers_force_refresher(session_timers_force_refresher)
+    {
+      if (typeof session_timers_force_refresher === 'boolean')
+      {
+        return session_timers_force_refresher;
       }
     },
 
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -358,7 +358,7 @@ module.exports = class RTCSession extends EventEmitter
     extraHeaders.push('Content-Type: application/sdp');
     if (this._sessionTimers.enabled)
     {
-      extraHeaders.push(`Session-Expires: ${this._sessionTimers.defaultExpires}`);
+      extraHeaders.push(`Session-Expires: ${this._sessionTimers.defaultExpires}${this._ua.configuration.session_timers_force_refresher ? ';refresher=uac' : ''}`);
     }
 
     this._request = new SIPMessage.InitialOutgoingInviteRequest(


ibc:
How would be the remote UAS behavior if the INVITE from JsSIP contains `;refresher=uac` in the `Session-Expires` header?

And assuming that the behavior is that the UAS will respect JsSIP as refresher: What about if JsSIP receives an incoming INVITE with `;refresher=uac`? I don't think this is covered by this PR, right?

kkozlik:
> How would be the remote UAS behavior if the INVITE from JsSIP contains ;refresher=uac in the Session-Expires header?

As per the RFC 4028:

> The UAC MAY include the refresher parameter with value 'uac' if it
> wants to perform the refreshes.  However, it is RECOMMENDED that the
> parameter be omitted so that it can be selected by the negotiation
> mechanisms described below
> 
>  ...
>
> After several INVITE/422 iterations, the request eventually arrives
> at the UAS.  The UAS can adjust the value of the session interval as
> if it were a proxy; when done, it places the final session interval
> into the Session-Expires header field in a 2xx response.  The
> Session-Expires header field also contains a 'refresher' parameter,
> which indicates who is doing the refreshing -- the UA that is
> currently the UAC, or the UA that is currently the UAS.  As the 2xx
> response travels back through the proxy chain, each proxy can observe
> the final session interval but can't change it.

So the UAS just see that UAC wants to do the refreshes so it let it do do them.

> And assuming that the behavior is that the UAS will respect JsSIP as refresher: What about if JsSIP receives an incoming INVITE with ;refresher=uac? I don't think this is covered by this PR, right?

No this is not covered by this PR. As stated in the PR description it is mainly for environments with client-side only initiated connection.

But as looking to `_handleSessionTimersInIncomingRequest()` function. In this case shall JsSIP reply with `refresher=uac` and let the other side do the refreshes.

ibc:
> > And assuming that the behavior is that the UAS will respect JsSIP as refresher: What about if JsSIP receives an incoming INVITE with ;refresher=uac? I don't think this is covered by this PR, right?
> 
> No this is not covered by this PR. As stated in the PR description it is mainly for environments with client-side only initiated connection.
> 
> But as looking to `_handleSessionTimersInIncomingRequest()` function. In this case shall JsSIP reply with `refresher=uac` and let the other side do the refreshes.

Can you implement that in this PR? Honestly we don't want to add incomplete features that are just valid for some scenarios.

@jmillan thoguths?

jmillan:
Yes, let's not open the door to uncovered scenarios please.

kkozlik:
Well @ibc how do you suggest the JsSIP should behave in in case it receive `refresher=uac?` in the INVITE? IMO the current behavior is correct. 

Or shall it force to be refresher even if it is in UAS role? That sounds weird especially when on both ends is JsSIP with `session_timers_force_refresher` set to true...

ibc:
> Well @ibc how do you suggest the JsSIP should behave in in case it receive `refresher=uac?` in the INVITE? IMO the current behavior is correct.

I don't really remember what the spec says in this regards (and definitely I have no time to check it now). I'm not telling that your proposed behavior is wrong, I'm just asking. Let me summarize my concerns:

- Does this PR cover all the scenarios (INVITE sent by JsSIP and INVITE received by JsSIP)?
- Should we do something different in the second scenario if the INVITE comes with `refresher=uac` or `refresher=uas` depending on whether our `session_timers_force_refresher` is true or false?

May be nothing else is needed, I'm just asking.

kkozlik:
Yea, I perfectly understand that you do not want accept incomplete feature, but IMO this should be fine with this PR. Maybe I did not understood you correctly, therefore I asked what do you miss.

To your concerns:
> Does this PR cover all the scenarios (INVITE sent by JsSIP and INVITE received by JsSIP)?

Yes. The PR modify only the behavior when INVITE sent by JsSIP. In the other case the behavior does not need to be changed.

> Should we do something different in the second scenario if the INVITE comes with refresher=uac or refresher=uas depending on whether our session_timers_force_refresher is true or false?

According to RFC, if incoming INVITE already contain `refresher` parameter, the UAS shall just copy it's value to the response. Which is what JsSIP already do. So no change is needed there.

Btw. shall I prepare some doc for the `session_timers_force_refresher` option? If so please give me some instructions. I have not found docs in the github sources. Neither any instructions for writing it.

ibc:
Everything is clear now, thanks.

Yes please, document the new field in https://github.com/versatica/jssip.net/blob/master/content/documentation/3.5.x/api/ua.html#L120 and make a PR in that project. Once we have it we'll merge both and publish a release.

Thanks.

kkozlik:
OK PR: https://github.com/versatica/jssip.net/pull/6 is created for the doc

PR #677, opened 2020-12-14
Enable custom From-Header for outbound calls

fabrykowski:
Enables setting a From-header different from the Contact-header for an outbound call.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -354,18 +354,25 @@ module.exports = class RTCSession extends EventEmitter
     }
     else
     {
-      if (options.from_uri)
+      if (options.fromUri)
       {
-        requestParams.from_uri = URI.parse(options.from_uri);
+        if (options.fromUri instanceof URI)
+        {
+          requestParams.from_uri = options.fromUri;
+        }
+        else
+        {
+          requestParams.from_uri = URI.parse(options.fromUri);
+        }
       }
-      if (options.from_display_name)
+      if (options.fromDisplayName)
       {
-        requestParams.from_display_name = options.from_display_name;
+        requestParams.from_display_name = options.fromDisplayName;
       }
     }
 
     // From-header has been modified
-    if (anonymous || options.from_uri)
+    if (anonymous || options.fromUri)
     {
       extraHeaders.push(`P-Preferred-Identity: ${this._ua.configuration.uri.toString()}`);
     }
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -350,7 +350,6 @@ module.exports = class RTCSession extends EventEmitter
       requestParams.from_display_name = 'Anonymous';
       requestParams.from_uri = new URI('sip', 'anonymous', 'anonymous.invalid');
 
-      extraHeaders.push(`P-Preferred-Identity: ${this._ua.configuration.uri.toString()}`);
       extraHeaders.push('Privacy: id');
     }
     else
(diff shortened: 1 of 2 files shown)


fabrykowski:
I'm willing to prepare the updated documentation as well, but I don't know where.

ibc:
What is the motivation for this?

fabrykowski:
For example multiple people working in an office with a central reception number. (e.g. support hotline)
All desk phones and VoIP-PSTN-carriers ignore the `display_name`-part of the From-header and display the username instead. So I need a way to send `From: <sip:+1234567890@example.com>;tag=xyz` in the INVITEs, despite REGISTERing as `From: <sip:desk-phone-123@example.com>;tag=abc`.

ibc:
For that, the `P-Preferred-Identity` header must be used as stated by the RFC:

- https://tools.ietf.org/html/rfc3325#section-9.2
- https://tools.ietf.org/html/rfc5876

Sending a random/unknown `From` header is not the way to go.

fabrykowski:
Unfortunately RFC 3325 is only a memo ("This memo provides information for the Internet community.  It does not specify an Internet standard of any kind.") and the PBXs in question only support the official RFC 3261 with the corresponding From-header ([8.1.1.3](https://tools.ietf.org/html/rfc3261#section-8.1.1.3) and [20.20](https://tools.ietf.org/html/rfc3261#section-20.20)).

ibc:
@jmillan thoughts?

ibc:
> Besides: if `anonymous` is set, it does exactly the same.

This is how anonymous call must be done according to the RFCs.

lib/RTCSession.js
```diff
@@ -353,6 +353,17 @@ module.exports = class RTCSession extends EventEmitter
       extraHeaders.push(`P-Preferred-Identity: ${this._ua.configuration.uri.toString()}`);
       extraHeaders.push('Privacy: id');
     }
+    else
+    {
+      if (options.from_uri)
+      {
+        requestParams.from_uri = URI.parse(options.from_uri);
```
ibc, line 360: `P-Preferred-Identity` is required if URI in From header does not honor the original one.

```suggestion
        requestParams.from_uri = URI.parse(options.from_uri);
        extraHeaders.push(`P-Preferred-Identity: ${this._ua.configuration.uri.toString()}`);
```

ibc (changes requested):

fabrykowski:
My precise use case:
I am developing a software for front-office employees, who call customers on behalf of other back-office employees. If the customer could not be reached, they should be able to call back the correct back-office employee.
I have run a test with three different PSTN endpoints and all of them ignored the contents of `P-Preferred-Identity`, while displaying the value from the `From` header.

jmillan:
Hi @fabrykowski, I see your scenario and concern. Here some changes to your proposal.

* Use `options.fromUserName` as **string** instead of providing a hole URI. We don't want to change the URI host.
* Slight code cleanup.

```js
    if (anonymous)
    {
      requestParams.from_display_name = 'Anonymous';
      requestParams.from_uri = new URI('sip', 'anonymous', 'anonymous.invalid');

      extraHeaders.push(`P-Preferred-Identity: ${this._ua.configuration.uri.toString()}`);
      extraHeaders.push('Privacy: id');
    }
    else if (options.fromUserName)
    {
      requestParams.from_uri = new URI('sip', options.fromUserName, this._ua.configuration.uri.host);

      extraHeaders.push(`P-Preferred-Identity: ${this._ua.configuration.uri.toString()}`);
    }

    if (options.fromDisplayName)
    {
      requestParams.from_display_name = options.fromDisplayName;
    }

```

fabrykowski:
I like the idea of `options.fromUserName`!

The standalone `if`-block for `fromDisplayName` allows to override the display name even if `anonymous` is set – I like this, because then the name can be localised.

ibc (approved):
I approve this, but it must first added to the doc website. Can you please do a PR also in there?

fabrykowski:
See versatica/jssip.net#8

fabrykowski:
Thanks!

jmillan:
New version released, thanks @fabrykowski!

PR #679, opened 2021-01-06
Ability to run JsSIP locally via a docker container

andrewjacksongamma:
This is to aid locally building, and reduce dependency installations on a a local machine.

Dockerfile, and documentation added

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,7 +1,13 @@
 ## What you need to build JsSIP
 
-You just need to have [Node.js](https://nodejs.org/) and [Git](https://git-scm.com/).
+This document details the steps for building JsSIP on your local machine. This
+will involve installing dependencies.
 
+You can run JsSIP via a [docker] container, where the dependencies for JsSIP are
+isolated to within that container. Details of how to do this are detailed
+[here](building-docker.md)
+
+You just need to have [Node.js](https://nodejs.org/) and [Git](https://git-scm.com/).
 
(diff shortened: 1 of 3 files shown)


ibc:
Thanks for this. However I don't like the idea of including Docker stuff into a simple JS library. We don't want to maintain a document about Docker usage and keep things as simple as possible. So thanks a lot but we prefer to not include this.

BTW you could create and publish a Docker image for JsSIP in Docker registry, right? No need to include it in JsSIP sources.

andrewjacksongamma:
I can understand the desire to keep things simple, and not wanting to maintain usage about Docker. I believe for the vast majority of the consumers of this library there will no requirement for using Docker. However for those who wish to contribution to the codebase having the ability to run the package in an isolated, recreatable, and consistent environment is in my opinon a profitable addition.

There is a potential danger when building thje library in it's current form that you will get differences due to people having different node versions. e.g. you may be running, compiling, testing, etc against node 12.x, whereas I could be running node 14.x, or even 15.x. There is no guarantee in the library itself that it will work with these versions of node, and the determination of  errors is placed on the consumer.

You are correct in that the Docker image can be published to a registtry. However therre still needs to be a underlying dockerfile to do so, and I believe the best place for such an file is to sit with the actual library.

ibc:
You can publish a jssip Docker image to the registry and provide with documented environment variables to use volumes, etc.

PR #693, opened 2021-03-12
Support 2 bytes long UTF-8 chars in the SDP

voidpp:
At this time, 2 bytes long UTF-8 characters are displayed as junk, becuase the `new Uint8Array` is splitting away them.

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -41,6 +41,8 @@ module.exports = class Transport
 
     this.status = C.STATUS_DISCONNECTED;
 
+    this.textDecoder = new TextDecoder('utf8');
+
     // Current socket.
     this.socket = null;
 
@@ -343,7 +345,7 @@ module.exports = class Transport
     {
       try
       {
-        data = String.fromCharCode.apply(null, new Uint8Array(data));
+        data = this.textDecoder.decode(data);
       }
       catch (evt)
       {


ibc:
Does it work in Node as well?

ibc:
Also, this should be done in the SDP parser rather than in the transport.

voidpp:
1. According to [this](https://nodejs.org/dist/latest-v15.x/docs/api/util.html#util_class_util_textdecoder),  TextDecoder is added in v8.3.0 (I tested with v12.20.0, and it's worked)
2. Well, I didn't want to refactor the code, only fix the actual issue.

ibc:
> 2\. Well, I didn't want to refactor the code, only fix the actual issue.

Then it must be properly fixed. Right now, your PR decodes the full SIP message while the only thing that must be decoded is the SDP. It must be done in `Parser.js` in `parseMessage()` function, here:

```js
  /* RFC3261 18.3.
   * If there are additional bytes in the transport packet
   * beyond the end of the body, they MUST be discarded.
   */
  if (message.hasHeader('content-length'))
  {
    const contentLength = message.getHeader('content-length');

    message.body = data.substr(bodyStart, contentLength);
  }
  else
  {
    message.body = data.substring(bodyStart);
  }
```

And we should do it based on the value of `Content-Type` SIP header. Example:

* `Content-Type: application/sdp; charset=utf-8`

It means that SDP is encoded in UTF-8. Same if `charset` is not given.

rshpount:
Per https://tools.ietf.org/html/rfc3261#section-7: SIP is a text-based protocol and uses the UTF-8 charset

Per https://tools.ietf.org/html/rfc4566#section-5:  An SDP session description is entirely textual using the ISO 10646 character set in UTF-8 encoding.  SDP field names and attribute names use only the US-ASCII subset of UTF-8, but textual fields and attribute values MAY use the full ISO 10646 character set.

You probably want to decode the entire message as UTF-8. Also, from my experience, you do not see 2 bytes or longer characters in the SDP, but they are common in the SIP headers (From/To header names). It would be illegal to use "Content-Type" header to overwrite the SDP charset.

ibc:
Thanks, Roman. Super detailed info. We'll manage this next week by adding needed tests.

rshpount:
@ibc When adding new tests, make sure to add tests with invalid UTF-8 characters in places like From or To name. They should be replaced with \uFFFD. Many SIP endpoints put non-UTF-8 characters (latin1) in the text fields, even though this is supposed to be only UTF-8. 

Also, if you support multipart bodies in INVITE/2XX responses, only SDP is UTF-8. Other parts can use other encodings. In robust parsers, you convert each SIP message line into UTF-8 and then parse. Once you reach the message body, you parse each content part separately. Only SDP as UTF-8. Other content parts can use other charsets or even be binary.

ibc:
@rshpount, I'm afraid we do not replace invalid UTF-8 chars with `\uFFFD` and doing that would require huge work. JavaScript is Unicode friendly and constraining it to be UTF-8 would be anti-natural. I've adding a test that parses a `NameAddr` with a display name containing a non UTF-8 char (`0xC0`). But it just behave as "expected". It get parsed as it is:

https://github.com/versatica/JsSIP/commit/b127103e348e7f3db90422fcd2340184cc58f508

Regarding multipart bodies, on it, let's see depending on how hard it is.

lib/Transport.js
```diff
@@ -343,7 +345,7 @@ module.exports = class Transport
     {
       try
       {
-        data = String.fromCharCode.apply(null, new Uint8Array(data));
+        data = this.textDecoder.decode(data);
```
ibc, line 348: Why is this just here for WebSocket binary messages?

ibc: ok, I assume this is just a problem in binary WebSocket messages.

ibc:
I'm gonna merge this PR as it is. The problem is, indeed, with WebSocket binary messages and the way we manage multi-byte chars in those **binary** messages (we split them in 2/3 single byte chars). So this PR is not fixing just something related to SDP but to the whole SIP message.

jmillan:
New release created. Thanks @voidpp !

lib/Transport.js
voidpp: Well, yes, maybe the summary of the PR is not entirely accurate.
(actually i had a 2 bytes long char in the SDP, not in a SIP header)

voidpp:
Thanks for the quick response!

lib/Transport.js
ibc: Yes, but we do the conversion of binary to text in the whole WebSocket binary message, before we can parse SIP headers and body. So it's ok as it is.

voidpp:
```
a=csrc:3755902497 cname:name@company.com
a=csrc:3755902497 msid:default
a=csrc:3755902497 name:Lajos Sánta
```

ibc:
> @ibc The TextDecoder is supposed to replace the invalid characters unless you set the "fatal" flag in options. Your test is not quite testing this. What you want is something that tests complete message decode, i.e., have a test where your test parses message with both valid and invalid UTF-8 in the message.

I don't know what to do here. Yes, the test does not test anything. It just receives a non-UTF-8 `display_name` in `NameAddr`and verifies that, once parsed, we have the very same value in the parsed object.

> The TextDecoder is supposed to replace the invalid characters unless you set the "fatal" flag in options.

```js
invalidUtf8 = Buffer.from([0xC0]);
// => <Buffer c0>

decoder = new TextDecoder('utf-8');
// => TextDecoder { encoding: 'utf-8', fatal: false, ignoreBOM: false }

decoder.decode(invalidUtf8);
// => '�'

// which is the same as if I do:

invalidUtf8.toString();
// => '�'
```

rshpount:
@voidpp Can you tell me what generates this? This is not a valid SDP since the name ssrc attribute is not defined.

ibc:
I'm using a `Buffer` because that's the only way to create a non-valid-UTF8 string (that's why I use `buffer.toString()` later). The thing is: I don't know what else should be done here.

I understand that Unicode other than valid UTF-8 should not be valid in SIP messages, but actually there is no real issues in accepting it when parsing received messages.

rshpount:
This is just a nit about the development process and can be ignored. 

Normally when you had a bug, you add a test to replicate the same condition. In this specific case, the entire SIP message was decoded using String.fromCharCode, which was replaced with TextDecoder. The test for this condition would be to decode the complete SIP message from a binary string using TextDecoder with ASCII, valid UTF-8, and invalid UTF-8 messages. I would even have a separate decodeData function implemented using TextDecoder and have unit tests just for that.

Once this is done, the unit test for NameAddress with UTF-8 is redundant.

ibc:
I see. The problem is that we do not test full SIP messages in current test units but just SIP message fragments.

voidpp:
> @voidpp Can you tell me what generates this? This is not a valid SDP since the name ssrc attribute is not defined.

It's an extension for internal use.

rshpount:
@voidpp FYI -- this is not standard compliant.

PR #697, opened 2021-03-23
Proposal for hook system in RTCSession

voidpp:
I need a solution to process the whole request in some special cases, so an event is not suitable for me, because I need to alter the request before it's sent out.

This proposed hook system do the job.
It's kinda basic implementation, but it's expandable.

What do you think about it?

### DIFF
--- a/lib/RTCSession.d.ts
+++ b/lib/RTCSession.d.ts
@@ -219,6 +219,13 @@ export interface RTCSessionEventMap {
   'peerconnection:setremotedescriptionfailed': AnyListener;
 }
 
+// hooks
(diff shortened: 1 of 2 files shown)


hthetiot:
I think EventEmmiter pattern would be more elegant than hooks like this, personnaly.

ibc:
Why doesn't an event work for you? The event may happen before the packet is sent so you can modify it.

voidpp:
Wait, EventEmitter's emit() is not async? So it's guaranteed all the event handlers will be called before we move on?

I assumed the opposite... (obviously)

ibc:
> EventEmitter's emit() is not async?

It's 100% sync.

voidpp:
Holy cow.

PR #701, opened 2021-04-06
Fixing values of UA.disconnected event

kkozlik:
The call of `this.ondisconnect()` function in `WebSocketInterface` calls `Transport._onDisconnect()` but this
function accept three arguments: `error`, `code` and `reason`. Not single data
object.

This causes that UA.disconnected event instead of having 4 simple
fields: `socket`, `error`, `code` and `reason` contain following object:
```
{
    socket: <socket instance>
    error: {
        socket: <socket instance>,
        error:  <valid data>,
        code:   <valid data>,
        reason: <valid data>
    },
    code: undefined,
    reason: undefined
}
```

### DIFF
--- a/lib/WebSocketInterface.js
+++ b/lib/WebSocketInterface.js
@@ -161,14 +161,7 @@ module.exports = class WebSocketInterface
       debug('WebSocket abrupt disconnection');
     }
 
-    const data = {
-      socket : this,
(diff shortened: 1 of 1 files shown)


jmillan:
Good catch @kkozlik, thanks!

jmillan:
applied in the new release

PR #709, opened 2021-05-13
Subscribe support

ikq:
With the proposed JsSIP modifications user can implement in his code client and server SUBSCRIBE/NOTIFY dialog.

Can be used different from REGISTER:
      - from/to user and domain
      -  different authentication (user/password/realm)

To working usage example, please see:
[https://github.com/ikq/subscribe_notify_test](https://github.com/ikq/subscribe_notify_test)


ikq:
I'll propose new version...

PR #712, opened 2021-05-27
Using other factors than only SIP URI user when trying to find a matching Contact in a Register 2xx response to prevent false matches

stefang42:
Currently Registrator.register() only uses the user part of a contact's SIP URI to find a match in the contacts received in a 2xx Register response from the server. This works fine with JsSIP's default contacts where the user part is a random token. However, this can lead to problems when the contact URI has been customized via UA config param contact_uri, which we do in our project. In our case the user part of the contact URI is set to the user part of the user's SIP credentials, so when the same user is registered on multiple devices with different clients the user part is identical for all bindings.

Consider the following example of a register:

```
common.js:108 Testapp:Softphone 2021-05-27T08:48:44.664Z (Debug): JsSIP:Transport sending message:REGISTER sip:dd146b51dfe7.us.dev.api.testhost.io SIP/2.0
Via: SIP/2.0/WSS ggvtu5dc2aam.invalid;branch=z9hG4bK3785002
Max-Forwards: 69
To: <sip:52aeb89a2913@dd146b51dfe7.us.dev.api.testhost.io>
From: "Stefan2Chrome" <sip:52aeb89a2913@dd146b51dfe7.us.dev.api.testhost.io>;tag=tebg4cft7a
Call-ID: 60b3mgqp1ec8na7aem71mu
CSeq: 2 REGISTER
Authorization: Digest algorithm=MD5, username="52aeb89a2913", realm="dd146b51dfe7.us.dev.api.testhost.io", nonce="YK9eGGCvXOy3Py8yoUlQOzdkMm8iNR/T", uri="sip:dd146b51dfe7.us.dev.api.testhost.io", response="eef582106d24b8edc96f325e81a9b85f"
Contact: <sip:52aeb89a2913@ggvtu5dc2aam.invalid;transport=ws>;+sip.ice;reg-id=1;+sip.instance="<urn:uuid:a3550190-ccef-4631-b06b-1f2382723ac1>";expires=600
Expires: 600
Allow: INVITE,ACK,CANCEL,BYE,UPDATE,MESSAGE,OPTIONS,REFER,INFO,NOTIFY
Supported: path,gruu,outbound
User-Agent: Softphone (1.6.0-beta.90/WebRTC/jsSIP 3.7.1/Chrome 90.0.4430.212/macOS 10.15.7)
Content-Length: 0

 +0ms +0ms
common.js:108 Testapp:Softphone 2021-05-27T08:48:44.664Z (Debug): JsSIP:WebSocketInterface send() +10ms +0ms
common.js:108 Testapp:Softphone 2021-05-27T08:48:44.862Z (Debug): JsSIP:WebSocketInterface received WebSocket message +198ms +198ms
common.js:108 Testapp:Softphone 2021-05-27T08:48:44.863Z (Debug): JsSIP:Transport received text message:SIP/2.0 200 OK
Via: SIP/2.0/WSS ggvtu5dc2aam.invalid;branch=z9hG4bK3785002;rport=53910;received=93.104.94.181
To: <sip:52aeb89a2913@dd146b51dfe7.us.dev.api.testhost.io>;tag=ddb25159a88af59214e4a3ffce8d28b7.150a
From: "Stefan2Chrome" <sip:52aeb89a2913@dd146b51dfe7.us.dev.api.testhost.io>;tag=tebg4cft7a
Call-ID: 60b3mgqp1ec8na7aem71mu
CSeq: 2 REGISTER
Contact: <sip:52aeb89a2913@192.168.178.59:63794;transport=tls;tinstance=64c9610a-5610-4965-8278-07fc4dc482c3>;expires=2525;received="sip:93.104.94.181:58033;transport=tls";+sip.instance="d9089d97-5fb5-4873-8248-5b667d9811ec", <sip:52aeb89a2913@ggvtu5dc2aam.invalid;transport=ws>;expires=600;received="sip:93.104.94.181:53910;transport=ws";+sip.instance="<urn:uuid:a3550190-ccef-4631-b06b-1f2382723ac1>";reg-id=1
Server:c3PR202d
Content-Length: 0
```

Here the 200 OK response contains two contacts in the Contact header, both have the same user part in the SIP URI, but different hosts. The current implementation of Registrator.register() therefore picks the first of the contacts, because it's the first one that matches the expected user part. However, it is a binding from another client with a much higher expires value, 2525 seconds instead of 600 seconds in the correct contact. This leads to an expiring registration on the server because JsSIP's registrationTimer has been configured using the incorrect 2525 value.

My change corrects that by also using the contact SIP URI's host and port parts in addition to the user parts when looking for a matching contact. This is also in line with the comparison rules defined in [section 19.1.4 of RFC 3261](https://datatracker.ietf.org/doc/html/rfc3261#section-19.1.4).

### DIFF
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -10,7 +10,7 @@ module.exports = class Registrator
 {
   constructor(ua, transport)
   {
-    const reg_id=1; // Force reg_id to 1.
+    this._reg_id=1; // Force reg_id to 1.
 
     this._ua = ua;
     this._transport = transport;
@@ -47,10 +47,10 @@ module.exports = class Registrator
     // Contents of the sip.instance Contact header parameter.
     this._sipInstance = '';
 
-    if (reg_id)
+    if (this._reg_id)
     {
       this._sipInstance = `"<urn:uuid:${this._ua.configuration.instance_id}>"`;
-      this._contact += `;reg-id=${reg_id}`;
+      this._contact += `;reg-id=${this._reg_id}`;
       this._contact += `;+sip.instance=${this._sipInstance}`;
     }
   }
@@ -171,12 +171,13 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
             // Get the Contact pointing to us and update the expires value accordingly.
             let contact = undefined;
 
-            // If we have a sip.instance parameter in our Contact header then try to
-            // Find a matching Contact using that.
-            if (this._sipInstance)
+            /* If we have a reg-id and a sip.instance parameter in our Contact header then try to
+               find a matching Contact using that. */
+            if (this._reg_id && this._sipInstance)
             {
               contact = contacts.find((element) => (
-                (this._sipInstance === element.getParam('+sip.instance'))
+                (this._sipInstance === element.getParam('+sip.instance')) &&
+                (this._reg_id === parseInt(element.getParam('reg-id')))
               ));
             }
 
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -166,7 +166,9 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
 
             // Get the Contact pointing to us and update the expires value accordingly.
             const contact = contacts.find((element) => (
-              element.uri.user === this._ua.contact.uri.user
+              (element.uri.user === this._ua.contact.uri.user) &&
+              (element.uri.host === this._ua.contact.uri.host) &&
+              (element.uri.port === this._ua.contact.uri.port)
             ));
 
             if (!contact)
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -10,7 +10,8 @@ module.exports = class Registrator
 {
   constructor(ua, transport)
   {
-    this._reg_id=1; // Force reg_id to 1.
+    // Force reg_id to 1.
+    this._reg_id = 1;
 
     this._ua = ua;
     this._transport = transport;
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -44,10 +44,14 @@ module.exports = class Registrator
     // Custom Contact header params for REGISTER and un-REGISTER.
     this._extraContactParams = '';
 
+    // Contents of the sip.instance Contact header parameter
+    this._sipInstance = '';
+
     if (reg_id)
     {
+      this._sipInstance = `"<urn:uuid:${this._ua.configuration.instance_id}>"`;
       this._contact += `;reg-id=${reg_id}`;
-      this._contact += `;+sip.instance="<urn:uuid:${this._ua.configuration.instance_id}>"`;
+      this._contact += `;+sip.instance=${this._sipInstance}`;
     }
   }
 
@@ -165,11 +169,24 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
               .reduce((a, b) => a.concat(b.parsed), []);
 
             // Get the Contact pointing to us and update the expires value accordingly.
-            const contact = contacts.find((element) => (
-              (element.uri.user === this._ua.contact.uri.user) &&
-              (element.uri.host === this._ua.contact.uri.host) &&
-              (element.uri.port === this._ua.contact.uri.port)
-            ));
+            let contact = undefined;
+
+            // if we have a sip.instance parameter in our Contact header then try to
+            // find a matching Contact using that.
+            if (this._sipInstance)
+            {
+              contact = contacts.find((element) => (
+                (this._sipInstance === element.getParam('+sip.instance'))
+              ));
+            }
+
+            // if no match was found using the sip.instance try comparing the URIs
+            if (!contact)
+            {
+              contact = contacts.find((element) => (
+                (element.uri.user === this._ua.contact.uri.user)
+              ));
+            }
 
             if (!contact)
             {
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -46,14 +46,10 @@ module.exports = class Registrator
     this._extraContactParams = '';
 
     // Contents of the sip.instance Contact header parameter.
-    this._sipInstance = '';
-
-    if (this._reg_id)
-    {
-      this._sipInstance = `"<urn:uuid:${this._ua.configuration.instance_id}>"`;
-      this._contact += `;reg-id=${this._reg_id}`;
-      this._contact += `;+sip.instance=${this._sipInstance}`;
-    }
+    this._sipInstance = `"<urn:uuid:${this._ua.configuration.instance_id}>"`;
+    
+    this._contact += `;reg-id=${this._reg_id}`;
+    this._contact += `;+sip.instance=${this._sipInstance}`;
   }
 
   get registered()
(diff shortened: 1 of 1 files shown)


ibc (changes requested):
If we do this change (that may work for you) all JsSIP instances connected to Asterisk or even Kamailio/OpenSIPS will fail since those servers typically change the host part of the Contact URI (they replace the private IP or whatever host part with the public received IP).

We cannot accept this change as it is because it will break things. I would accept a configurable new UA setting (`match_host_port_in_contact_uri` which defaults to `false`).

stefang42:
Ok, understood, I wasn't aware of that. That's strange however, doesn't that violate RFC 3261, section 19.1.4 then?

>          For two URIs to be equal, the user, password, host, and port
>          components must match.
> 

Another idea might be to also consider the sip.instance contact parameter instead, if present. This should be unique and is hopefully not changed by the server. However, I'm not sure if that can be accessed there and would not violate RFC 3261 as it is not part of the contact SIP URI. What do you think?

ibc:
Matching Contact URI based on `sip.instance` (**if present**) is probably the way to go. It's obviously not defined in RFC 3261 which is pre-feature, but should be defined in some other RFC.

stefang42:
I have just committed another proposal. This time I check if we are using a sip.instance parameter in our Register requests and if so then I try to find a matching Contact in the 2xx response using that. If that fails then the old method of matching SIP URI user is used. Is that acceptable?

lib/Registrator.js
```diff
@@ -44,10 +44,14 @@ module.exports = class Registrator
     // Custom Contact header params for REGISTER and un-REGISTER.
     this._extraContactParams = '';
 
+    // Contents of the sip.instance Contact header parameter
```
ibc, line 47:
```suggestion
    // Contents of the sip.instance Contact header parameter.
```

lib/Registrator.js
```diff
@@ -165,9 +169,24 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
               .reduce((a, b) => a.concat(b.parsed), []);
 
             // Get the Contact pointing to us and update the expires value accordingly.
-            const contact = contacts.find((element) => (
-              element.uri.user === this._ua.contact.uri.user
-            ));
+            let contact = undefined;
+
+            // if we have a sip.instance parameter in our Contact header then try to
```
ibc, line 174:
```suggestion
            // If we have a sip.instance parameter in our Contact header then try to
```

lib/Registrator.js
```diff
@@ -165,9 +169,24 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
...
+            // find a matching Contact using that.
```
ibc, line 175:
```suggestion
            // Find a matching Contact using that.
```

lib/Registrator.js
```diff
@@ -165,9 +169,24 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
...
+            if (this._sipInstance)
+            {
+              contact = contacts.find((element) => (
+                (this._sipInstance === element.getParam('+sip.instance'))
+              ));
+            }
+
+            // if no match was found using the sip.instance try comparing the URIs
```
ibc, line 183:
```suggestion
            // If no match was found using the sip.instance try comparing the URIs.
```

ibc (changes requested):
Some cosmetic changes requested but LGTM. @jmillan do we miss something here?

stefang42:
Thanks, I have committed your suggestions.

rshpount:
The sip.instance Contact header parameter is defined in rfc5626, a recommended solution for handling SIP registrations from behind NAT.

I think it would be cleaner if you match the contact based on both sip.instance and URI user part at the same time.

ibc:
> I think it would be cleaner if you match the contact based on both sip.instance and URI user part at the same time.

Why would that better than just matching the full sip.instance?

rshpount:
> > I think it would be cleaner if you match the contact based on both sip.instance and URI user part at the same time.
> 
> Why would that better than just matching the full sip.instance?

Sorry, I meant to say that you have to compare sip.instance and reg-id. The value of sip.instance is not unique. Multiple registrations are supposed to be created using the same sip.instance and different reg-id values to create redundant flows between the server and the client.

ibc:
Good point.

stefang42:
Thanks for the valuable input! I have added a check for the reg-id with the latest commit, I hope this is acceptable now.

stefang42:
@ibc @jmillan Do you need me to make any more changes? Or do you think this is ready to be merged?

lib/Registrator.js
```diff
@@ -10,7 +10,7 @@ module.exports = class Registrator
 {
   constructor(ua, transport)
   {
-    const reg_id=1; // Force reg_id to 1.
+    this._reg_id=1; // Force reg_id to 1.
```
jmillan, line 13:
```suggestion
    // Force reg_id to 1.
    this._reg_id = 1;
```

lib/Registrator.js
```diff
@@ -44,10 +44,14 @@ module.exports = class Registrator
     // Custom Contact header params for REGISTER and un-REGISTER.
     this._extraContactParams = '';
 
-    if (reg_id)
+    // Contents of the sip.instance Contact header parameter.
+    this._sipInstance = '';
+
+    if (this._reg_id)
```
jmillan, line 50: Let's remove this conditional, since we always setting it on the top of the constructor. (I know this was already this way).

lib/Registrator.js
```diff
@@ -165,9 +169,25 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
...
 
             // Get the Contact pointing to us and update the expires value accordingly.
-            const contact = contacts.find((element) => (
-              element.uri.user === this._ua.contact.uri.user
-            ));
+            let contact = undefined;
+
+            /* If we have a reg-id and a sip.instance parameter in our Contact header then try to
+               find a matching Contact using that. */
+            if (this._reg_id && this._sipInstance)
```
jmillan, line 176: This condition is not needed because both members are always set.

jmillan (commented):
Apart from few little changes requested, I see it OK.

lib/Registrator.js
stefang42: Agreed, I was just leaving it in there because it was already there before.

lib/Registrator.js
stefang42: I just had those checks there in case someone changed the initialization in the constructor, but I agree, they are not really needed.

stefang42:
@jmillan Thanks for reviewing, I have added your suggestions with the latest commits.

lib/Registrator.js
```diff
@@ -164,11 +165,22 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
             const contacts = response.headers['Contact']
               .reduce((a, b) => a.concat(b.parsed), []);
 
-            // Get the Contact pointing to us and update the expires value accordingly.
-            const contact = contacts.find((element) => (
-              element.uri.user === this._ua.contact.uri.user
+            /* Get the Contact pointing to us and update the expires value accordingly.
+               If we have a reg-id and a sip.instance parameter in our Contact header then try to
+               find a matching Contact using that. */
```
ibc, line 170: I know this is cosmetic but let's follow code convention:

```suggestion
            // Get the Contact pointing to us and update the expires value accordingly.
            // If we have a reg-id and a sip.instance parameter in our Contact header then try
            // to find a matching Contact using that.
```

ibc (approved):
Just a super cosmetic change requested. Thanks a lot.

stefang42:
@ibc Ok, done. I was using /* */ because multiline comments using // tend to be placed weirdly in the generated lib-es5 code.

lib/Registrator.js
```diff
@@ -165,10 +166,21 @@ ${this._contact};expires=${this._expires}${this._extraContactParams}`);
               .reduce((a, b) => a.concat(b.parsed), []);
 
             // Get the Contact pointing to us and update the expires value accordingly.
-            const contact = contacts.find((element) => (
-              element.uri.user === this._ua.contact.uri.user
+            // If we have a reg-id and a sip.instance parameter in our Contact header then try
+            // to find a matching Contact using that.
```
jmillan, line 170: Cosmetic, this does not apply anymore because reg-id and sip-instance are always there.

` Try to find a matching Contact using sip.instance and reg-id.`

stefang42: Ok, done.

jmillan:
Thanks so much @stefang42!

jmillan (approved):

stefang42:
Thanks again everyone for the valuable input and for accepting this change, much appreciated!

jmillan:
JsSIP version 3.7.6 has been released.

Issue #713, opened 2021-05-27
New Feature: allow to override debugerror


lamuertepeluda:
## Use Case
I need to intercept the logs from both errors and standard debug log messages, e.g. for sending them to a backend. 

Logging standard messages is easy by overriding the debug.log function, e.g. as shown here: https://groups.google.com/g/jssip/c/5LWejj-N_sU, i.e.

```javascript
JsSIP.debug.log = (...args) => {
    const logString = args[0].replace(/%c/g, '');  // Remove coloring syntax used by debug tool.
    logger(logString);
};
```

Logging `debugerror` is however impossible, since it's internally defined and bound to console.warn() in an isolated scope.  
E.g. [like in this commit](https://github.com/versatica/JsSIP/commit/8b4e547c731c5ec92cbe44fd320c927104dddf02), or [here in the sources](https://github.com/versatica/JsSIP/blob/master/lib/WebSocketInterface.js#L5)

```javascript
var debugerror = require('debug')('JsSIP:ERROR:Socket');
debugerror.log = console.warn.bind(console);
```

Because of this implementation, [exposing the debug module as it's done here](https://github.com/versatica/JsSIP/blob/master/lib/JsSIP.js#L27) is not enough.

## Proposal

Have an option for overriding the default logger(s), with default to the current behaviour. 

I can try a PR if you like this idea.

jmillan:
> Have an option for overriding the default logger(s), with default to the current behaviour.

Feel free to create a PR.

PR #714, opened 2021-05-31
Intercept debug

lamuertepeluda:
This PR should solve #713 

*This PR is a drop-in replacement* of the log calls mechanism. It does not change the user API. I only replaced the inner `require('debug')` calls with the require calls to the new `JsSIPDebug` module.

I wrapped the debug module in a function which adds two module-level setters that will setup a custom default log function and a custom default error log function.

```js
// a custom logger which prepends some stuff
const myLogger = (...args) => console.log.apply(console, ['🤪 Cool Custom log 🤪', ...args]);

// a custom error logger which prepends some stuff and writers onn console.error rather than console.warn
const myErrorLogger = (...args) => console.error.apply(console, ['💀 There were Errors 💀', ...args])

// set myLogger as default logger
JsSIP.debug.setLogger(myLogger)

// set myErrorLogger as default error logger
JsSIP.debug.setErrorLogger(myErrorLogger)
```

In this way, the user is able to override the logging methods and it is possible to transform the log messages, or send them to a remote server, for instance.

If you don't setup any custom logger, the library will just keep the current behaviour.

This PR also makes the `log` function not writable, therefore the setters are the only way to change the log behaviour, i.e. the following:

```js
JsSIP.debug.log = (...args) => {
    const logString = args[0].replace(/%c/g, '');  // Remove coloring syntax used by debug tool.
    logger(logString);
};
```

will not work anymore. You will have to `JsSIP.debug.setLogger(...)` to achieve the same result. This should not be an issue, since this is an undocumented feature.

If approved, this feature should be added to the JsSIP documentation.

### Working Example

The following is an example with the new setters on Node

jssip_setup

The following is the output coming from the above setup

Output
jssip_output


jmillan:
Hi @lamuertepeluda,

@ibc and I use this kind of Logger class in other projects. Let me handle this in a separate PR since we want to do it the same way since it does not make sense making it differently.

Thanks very much for pushing this.

NOTE: I'll keep this PR open until I create the new one today or tomorrow.

jmillan:
@lamuertepeluda, by the way, what a nick name :-)

lamuertepeluda:
@jmillan ok, feel free to pick stuff from my implementation, if you find it helpful in any way, then! And let me know if I can help.

The nickname is a reference from a funny old South Park episode 😆

michelepra:
my 2¢ on this.
This also can be done with debug module. I'v do this some year ago

```typescript
import { Log, LogLevel } from "../lib/Logger";
import * as JsSIP from "jssip";

declare module "debug" {
    interface Debug {
        namespace: string;
    }
}

const jssipDebugEnable = "*:JsSIP:Transport,*:JsSIP:RTCSession";
const jsSipLogger = new Map<String, Log>();
const jssipLogToLogLevelMapper = new Map<String, LogLevel>();
jssipLogToLogLevelMapper.set("ERROR", LogLevel.Error);
jssipLogToLogLevelMapper.set("Dialog", LogLevel.Verbose);
jssipLogToLogLevelMapper.set("Transport", LogLevel.Debug);
jssipLogToLogLevelMapper.set("RTCSession", LogLevel.Info);

JsSIP.debug.enable(jssipDebugEnable);
JsSIP.debug.formatArgs = function (args: any[]) {
    return args[0];
};
JsSIP.debug.log = function () {
    let logger = jsSipLogger.get(this.namespace);
    if (!logger) {
        logger = Log.of(this.namespace);
        jsSipLogger.set(this.namespace, logger);
    }

    if (this.namespace.endsWith(":RTCSession") && arguments[0]) {
        const msg: string = arguments[0];
        if (msg.startsWith('emit "sending"')) {
            arguments[0] = 'emit "sending"';
        }
    }

    const argv: [LogLevel, string, any[]] = [jssipLogToLogLevelMapper.get(this.namespace.substring(this.namespace.lastIndexOf(":") + 1)) || LogLevel.Debug].concat(Array.from(arguments)) as any;
    logger.log.apply(logger, argv);
};

```

This code is only one example
If you have plan to change log stack considering to add prefix on every logline, if you run more than one instance of UA is unclear how to analyze log

ibc:
There is a common confusion here. JsSIP logs (as those logs generate for example by Nose express library which also uses debug module) are not logs for diagnosing what's going on in a production environment. Those are logs for development.

michelepra:
> There is a common confusion here. JsSIP logs (as those logs generate for example by Nose express library which also uses debug module) are not logs for diagnosing what's going on in a production environment. Those are logs for development.

so how to log for diagnosing  in production?? i must handle all event and log by myself (very tedious)?

lamuertepeluda:
@ibc I see what you mean, but I don't see why we shouldn't we allowed to tap in the logging system in order to send them to a remote diagnosing tool, for instance.

Furthermore I also think that sometimes it is quicker to diagnose problems like this, as @michelepra  pointed out.

About the implementation: I think there are several ways to achieve this. I also suggested some little changes to #715 that would fit just fine. Any ETA for it to be merged?

lamuertepeluda:
@ibc in addition to the the extra effort required for manually adding all the error handlers, there are also some points in the code where the exceptions being thrown are not meaningful, but logs are, e.g. look at this line https://github.com/versatica/JsSIP/blob/76756f40563e9750fe5dc095f048c337021f9f40/lib/Socket.js#L46 and at the following line https://github.com/versatica/JsSIP/blob/76756f40563e9750fe5dc095f048c337021f9f40/lib/Socket.js#L47

Therefore in cases such as these intercepting errors through logs is also more helpful for diagnosis

ibc:
Those empty `Error` with no message are clearly something that should be fixed.

jmillan:
Definitely throwing empty errors is a bug

lamuertepeluda:
@jmillan @ibc I totally agree. Maybe better put that one in a different issue.
I also think that overriding the logs alone is not enough, and the error events and the exception handling should also be used when possible, however this is a use case that some people (including me) have, as I have seen also on Google Forum. Therefore if there is a way of implementing this without changing the default behaviour, I don't personally see any particolar reason not to.

jmillan:
No more empty messages are thrown in JsSIP.

lamuertepeluda:
@ibc , @jmillan I added here the changes in your master branch (which just merged #715 ) plus:
- [the additions to Logger.js I was suggesting](https://github.com/versatica/JsSIP/pull/715#issuecomment-852013352) for closing this issue
- the type definitions for `Logger.js` (`Logger.d.ts`, which were completely missing)
- integrated Logger also in `JsSIP.js` and `JsSIP.d.ts` *keeping backwards compatibility* with `JsSIP.debug.enable()` and `JsSIP.debug.disable()`

Let me know if this PR is now OK to be merged, or if there is anything else that I can do about it.

jmillan:
Sorry @lamuertepeluda,

We've been talking internally and don't see this PR something to be added to the library. The logs are a development tool only. The library offers an API composed by classes, methods,  events and exceptions, and those should be used by the application or service, not the logs.

ibc:
A different rationale is:

- This PR tries to make JsSIP a Logger factory.
- I'm pretty sure you are not requesting the same to other JS libs that you use (React, Vue, etc)?
- JsSIP debugs are intended to provide us with debugging during development. They are not "app logs" and we don't want they to become that.
- If we accept this PR, from now one JsSIP is (also) a logger factory and something that was mostly private (debugging) would be now publicly exposed.

lamuertepeluda:
@ibc I see, but I disagree. I don't want to appear argumentative, but I need to confute: 

> This PR tries to make JsSIP a Logger factory.

not really, it just makes it possible to replace the default logging functions with something else, including sending the logs to a server for analysis or not printing anything on some user-defined conditions (e.g. in a production environment).

> I'm pretty sure you are not requesting the same to other JS libs that you use (React, Vue, etc)? 

nope, but they have completely different purposes and logging mechanisms, and they do not fit my use case at all. React, which I know quite well, has some global handlers (error boundaries) where you can enter your own reporting logic, but still I don't see how this comparison would fit.

> JsSIP debugs are intended to provide us with debugging during development. They are not "app logs" and we don't want they to become that.

If this were completely true, why do we get the debug module exposed also in the production release (`dist/jssip.min.js`)? It wouldn't make sense to me. The fact is, many people use the debug module mainly to debug their app, rather than debugging the library itself. Sometimes, like in my use case, your app users are not technically skilled enough to start the debugger (or even open the dev tools). But getting the sequence of log messages is often enough for troubleshooting.

> If we accept this PR, from now one JsSIP is (also) a logger factory and something that was mostly private (debugging) would be now publicly exposed.

I personally don't see any big issue here since the library is FOSS and debug logs are not hidden (see my previous reply about them being available in the production library). You would expose just another API to intercept log messages, that users can decide whether to use or not, without interfering with your current API behavior in any way.

That said, of course the library and thus the final decision is yours.

ibc:
> If this were completely true, why do we get the debug module exposed also in the production release?

I agree this was our failure since it opened the door to this kind of feature requests. Anyway, the only reason to expose the `debug` module is for users to be able to set the debug namespace programatically (instead of relying on the `debug` entry in `localStorage`).

In the other side, we plan to do a completely new version of JsSIP which will be completely different. It won't deal with the WebRTC stack anymore but just with SIP signaling, and it will expose a completely different (and more flexible) API. So honestly, we want to avoid huge changes in current version in the meantime until that happens.

lamuertepeluda:
@ibc

> Anyway, the only reason to expose the debug module is for users to be able to set the debug namespace programatically (instead of relying on the debug entry in localStorage).

I understand that the previous one was perhaps not the best design pattern and would not really work on Node.js (not without polyfills and other dirty workarounds). But I can really see a few useful use cases for doing this programmatically while getting the logs only in the dev tools.

> we want to avoid huge changes in current version in the meantime until that happens.

I tried to make the changes as little as possible. The API wouldn't _change_, but this PR would only have _added_ 3 optional static methods to your logger class for overriding the default debugger _output_. It wouldn't mess in any way in how the library is currently working, or designed. Think of this like redirecting the standard output or the standard error of a program.

Therefore it seems to me that you are trying to put this discussion more on a kind of a philosophical level, but on the practical level I believe it wouldn't change really that much for JsSIP users. I'd say almost nothing. It could only make easier remote app troubleshooting.

So if you really are planning a huge refactor (say, a V4) of the library, you would also be free to discard any existing API, including the debug module (at least to a certain level). I was suggesting only a minor version update.
But once again, it's up to you.

Issue #689, opened 2021-02-26
JsSIP 3.7.1 doesn't allow sending DTMF during EarlyMedia


divisumma18:
JsSIP 3.7.1 seems to have the same issue described here:
https://twinklephone.yahoogroups.narkive.com/aXB3Vk5Q/twinkle-doesn-t-allow-sending-dtmf-during-earlymedia

TLDR: Sending a DTMF tone with `JsSIP.RTCSession.sendDTMF()` after a "183 Session Progress" and before a "SIP/2.0 200 OK" should be possibile, but JsSIP raises an "INVALID_STATE_ERROR: Invalid status: 2" error instead.

jmillan:
Wow, your pointing a 13 year old forum content.

Can you try this patch:

```diff
diff --git a/lib/RTCSession.js b/lib/RTCSession.js
index 7003c25..dcb41c3 100644
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -962,7 +962,11 @@ module.exports = class RTCSession extends EventEmitter
     }

     // Check Session Status.
-    if (this._status !== C.STATUS_CONFIRMED && this._status !== C.STATUS_WAITING_FOR_ACK)
+    if (
+      this._status !== C.STATUS_CONFIRMED &&
+      this._status !== C.STATUS_WAITING_FOR_ACK &&
+      this._status !== C.STATUS_1XX_RECEIVED
+    )
     {
       throw new Exceptions.InvalidStateError(this._status);
     }
```

Please also try what would happen if you call `sendDtmfs()` after receiving a `100` but before receiving `18X`.

divisumma18:
> Wow, your pointing a 13 year old forum content.

I thought an old post from Iñaki would still be pretty convincing! :wink:

> Can you try this patch:

I tried that and similarly patched DTMF.js

```diff
diff --git a/lib/RTCSession/DTMF.js b/lib/RTCSession/DTMF.js
index d27b4ac..5665c40 100644
--- a/lib/RTCSession/DTMF.js
+++ b/lib/RTCSession/DTMF.js
@@ -48,9 +48,9 @@ module.exports = class DTMF extends EventEmitter
     this._direction = 'outgoing';
 
     // Check RTCSession Status.
-    if (this._session.status !== this._session.C.STATUS_CONFIRMED &&
-      this._session.status !== this._session.C.STATUS_WAITING_FOR_ACK)
-    {
+    if (
[...]

PR #717, opened 2021-06-15
Fix invalid state error on DTMF send

BlindChickens:
Sending a DTMF tone after a `183 Session Progress` for example would raise an `INVALID_STATE_ERROR`. This PR fixes that.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -961,7 +961,11 @@ module.exports = class RTCSession extends EventEmitter
     }
 
     // Check Session Status.
-    if (this._status !== C.STATUS_CONFIRMED && this._status !== C.STATUS_WAITING_FOR_ACK)
+    if (
+      this._status !== C.STATUS_CONFIRMED &&
+      this._status !== C.STATUS_WAITING_FOR_ACK &&
+      this._status !== C.STATUS_1XX_RECEIVED
+      )
     {
       throw new Exceptions.InvalidStateError(this._status);
     }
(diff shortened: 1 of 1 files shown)


ibc (approved):

jmillan (approved):

jmillan:
Hi,

Reverted due to this https://github.com/versatica/JsSIP/issues/689

BlindChickens:
Hi, I cannot exactly deduce what extra conditions I need to test on. If you can clarify I can test.
What I can say is that before the change it does not work for me right after a `183 Session Progress`, and after the change it works anytime from `183 Session Progress` to `200 OK`.

jmillan:
I guess you are sending DTMFs of type JsSIP_C.DTMF_TRANSPORT.RFC2833, right?

ibc:
Can I know more? It should indeed be possible to send DTMF (in band or via INFO) once we have received a 18X.

jmillan:
> Can I know more? It should indeed be possible to send DTMF (in band or via INFO) once we have received a 18X.

That's why I ask. I want to confirm this, because INFO DTMFs for sure don't work.

ibc:
Why not?

ibc:
It should be possible to send in-dialog messages as far as the dialog is created, and that happens once we get a 18X with Contact header.

jmillan:
Due to the problem posted here #689. Yes, this is a bug in the library.

ibc:
Si I think such a bug must be fixed, yes, but no real reason to revert this PR.

BlindChickens:
> I guess you are sending DTMFs of type JsSIP_C.DTMF_TRANSPORT.RFC2833, right?

Yes indeed I am.

BlindChickens:
If I send the DTMF of type `INFO`, there is no error on my side, but also no corresponding sound on the other end indicating that it worked. I don't know what the differences are here, but I am going to assume our service provider does not allow for the `INFO` type DTMF's. I can see the websocket messages going out when I send them, but no sound on the other end.

jmillan:
> Si I think such a bug must be fixed, yes, but no real reason to revert this PR.

That's a point.

We are now preventing RFC2833 DTMFs from being sent. Ok, let's merge it.

jmillan:
Changes added to the new release.

PR #720, opened 2021-06-22
Allow overriding the Referred-By header on REFER

### LINKED ISSUE #719: New Feature: Allow overriding Referred-By header
In our use case, a UA registers with a user ID, say `12345@host` and may receive calls on multiple DIDs. Our Asterisk ARI application delivers calls by dialling `12345@host` and setting the target DID in the `To` header.

When the call is answered and REFER'd later on, the `Referred-By` header is always `12345@host` (since that's how we are registered with the UA), but we need this to be the dialed DID@host. We need flexibility to override this header, its logic is currently hardcoded in #572.

jp1987:
Fixes #719

### DIFF
--- a/lib/RTCSession/ReferSubscriber.js
+++ b/lib/RTCSession/ReferSubscriber.js
@@ -55,7 +55,8 @@ module.exports = class ReferSubscriber extends EventEmitter
     extraHeaders.push(referTo);
 
     // Referred-By header field (if not already present).
-    if (!extraHeaders.some(header => header.startsWith('Referred-By:'))) {
+    if (!extraHeaders.some(header => header.startsWith('Referred-By:'))) 
+    {
       const referredBy = `Referred-By: <${this._session._ua._configuration.uri._scheme}:${this._session._ua._configuration.uri._user}@${this._session._ua._configuration.uri._host}>`;
       extraHeaders.push(referredBy);
     }
--- a/lib/RTCSession/ReferSubscriber.js
+++ b/lib/RTCSession/ReferSubscriber.js
@@ -58,6 +58,7 @@ module.exports = class ReferSubscriber extends EventEmitter
     if (!extraHeaders.some(header => header.startsWith('Referred-By:'))) 
     {
       const referredBy = `Referred-By: <${this._session._ua._configuration.uri._scheme}:${this._session._ua._configuration.uri._user}@${this._session._ua._configuration.uri._host}>`;
+
       extraHeaders.push(referredBy);
     }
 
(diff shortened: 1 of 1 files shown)


lib/RTCSession/ReferSubscriber.js
```diff
@@ -54,10 +54,12 @@ module.exports = class ReferSubscriber extends EventEmitter
...
 
-    // Referred-By header field.
-    const referredBy = `Referred-By: <${this._session._ua._configuration.uri._scheme}:${this._session._ua._configuration.uri._user}@${this._session._ua._configuration.uri._host}>`;
-
-    extraHeaders.push(referredBy);
+    // Referred-By header field (if not already present).
+    if (!extraHeaders.some(header => header.startsWith('Referred-By:'))) 
+    {
+      const referredBy = `Referred-By: <${this._session._ua._configuration.uri._scheme}:${this._session._ua._configuration.uri._user}@${this._session._ua._configuration.uri._host}>`;
+      extraHeaders.push(referredBy);
```
jmillan, line 61:
```suggestion

      extraHeaders.push(referredBy);
```

jmillan:
Thanks @jp1987!

jmillan:
BTW, no new minor release has been created for this change.

I've just added this info in the current documentation: https://jssip.net/documentation/3.7.x/api/session/#method_refer

jp1987:
> BTW, no new minor release has been created for this change.
> 
> I've just added this info in the current documentation: https://jssip.net/documentation/3.7.x/api/session/#method_refer

Thank you @jmillan!

jp1987:
> BTW, no new minor release has been created for this change.
> 
> I've just added this info in the current documentation: https://jssip.net/documentation/3.7.x/api/session/#method_refer

You may want to change "Referred-by" to "Referred-By" in the documentation, it's case sensitive.

lib/RTCSession/ReferSubscriber.js
```diff
@@ -54,10 +54,13 @@ module.exports = class ReferSubscriber extends EventEmitter
 
     extraHeaders.push(referTo);
 
-    // Referred-By header field.
-    const referredBy = `Referred-By: <${this._session._ua._configuration.uri._scheme}:${this._session._ua._configuration.uri._user}@${this._session._ua._configuration.uri._host}>`;
+    // Referred-By header field (if not already present).
+    if (!extraHeaders.some(header => header.startsWith('Referred-By:')))
```
ibc, line 58: This is wrong. It should not be case sensitive. SIP headers are not case sensitive.

jp1987: Then it should be `header.toLowerCase().startsWith('referred-by:')`. New PR or can you guys fix this in master?

jmillan: > SIP headers are not case sensitive

Absolutely right.

Fixed in new release.

PR #721, opened 2021-06-27
Allow sending and responding to SIP OPTIONS messages

simonlindberg:
Useful for checking capabilities and the status of remote endpoints.

Notes:
- `Options.js` only has some minor differences from `Messages.js` (default `contentType`, no requirement on a body), but didn't feel comfortable generalizing
- Also reusing `_applicants` and `destroyMessage,` previously only used for `SIP MESSAGE` messages.

### DIFF
--- a/lib/Options.js
+++ b/lib/Options.js
@@ -1,0 +1,280 @@
+const EventEmitter = require('events').EventEmitter;
+const Logger = require('./Logger');
+const JsSIP_C = require('./Constants');
+const SIPMessage = require('./SIPMessage');
+const Utils = require('./Utils');
+const RequestSender = require('./RequestSender');
+const Exceptions = require('./Exceptions');
+
+const logger = new Logger('Options');
+
+module.exports = class Options extends EventEmitter
+{
+  constructor(ua)
+  {
+    super();
+
+    this._ua = ua;
+    this._request = null;
+    this._closed = false;
+
+    this._direction = null;
+    this._local_identity = null;
+    this._remote_identity = null;
+
+    // Whether an incoming message has been replied.
+    this._is_replied = false;
+
+    // Custom message empty object for high level use.
+    this._data = {};
+  }
+
+  get direction()
+  {
+    return this._direction;
+  }
+
+  get local_identity()
+  {
+    return this._local_identity;
+  }
+
+  get remote_identity()
+  {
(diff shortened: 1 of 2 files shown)


jmillan:
Hi @simonlindberg,

We know the rationale behind this, but Is this useful in real life? We have never seen this requirement.

simonlindberg:
Hi @jmillan , thank you for taking the time!

We would only use the JsSIP _sending_ `OPTIONS` part of the PR.
We have a browser based SIP-phone, based on JsSIP. Once a client has connected and registered using JsSIP, the client would start sending `OPTIONS` messages every X seconds.

For us the benefits would be twofold:
- **Faster client reconnection**; currently it would take minutes for a reconnection if a client abruptly changes network. It would take until the next `REGISTER` is sent for the socket to detect that the network is down, since [browsers doesn't](https://www.w3.org/Bugs/Public/show_bug.cgi?id=13104) [send websocket pings](https://bugs.chromium.org/p/chromium/issues/detail?id=706002) (and the fact that [onClose is slow to trigger](https://bugs.chromium.org/p/chromium/issues/detail?id=426798) doesn't help). With the periodic `OPTIONS` being sent and used as keepalive ping, the client would be aware and can reconnect within seconds instead.
- **Traffic shaping**; if the server that the client is connected to is to be taken down for maintence or similair a negative reponse to the `OPTIONS` could prompt a controlled reconnection to a different set of servers.

ibc:
I'm fine with this PR. I'd just also request adding the corresponding documentation. @simonlindberg may you please write a separate PR in https://github.com/versatica/jssip.net with the new API? It should mostly be copy&paste from MESSAGE API.

simonlindberg:
> I'm fine with this PR. I'd just also request adding the corresponding documentation. @simonlindberg may you please write a separate PR in https://github.com/versatica/jssip.net with the new API? It should mostly be copy&paste from MESSAGE API.

I’ll get to it!

simonlindberg:
Docuemntation changes up for review: https://github.com/versatica/jssip.net/pull/9!

jmillan:
Thanks @simonlindberg!

jmillan:
New JsSIP release 3.8.0 is out

simonlindberg:
Thank you!

Issue #732, opened 2021-09-15
Second INVITE gets sent if we put session on hold right before reINVITE to refresh session


makstheimba:
Hi!
While using your library we saw that some of our calls were being terminated due to second INVITE being sent during another INVITE being in progress. After closer inspection of packets we saw that that happens if we put session on hold right *before* reINVITE for session refresh is sent. We confirmed this issue and we able to easily and reliably reproduce it by adding a call to `hold()` the call milliseconds *before* reINVITE for session refresh is sent.

That doesn't happen however if we try to put session on hold right *after* reINVITE for session refresh is sent. Because inside the `hold()` call there is a check if there is currently another INVITE/UPDATE transaction in progress.

I think fixing this issue should be really easy - we just need to check `_isReadyToReOffer` during execution of session timer, before sending reINVITE to prolong session.
```
// RTCSession.js
...
    if (this._sessionTimers.refresher)
    {
      this._sessionTimers.timer = setTimeout(() =>
      {
        if (this._status === C.STATUS_TERMINATED) { return; }

        if (!this._isReadyToReOffer()) { return; } // <================== I've added this line

        logger.debug('runSessionTimer() | sending session refresh request');

        if (this._sessionTimers.refreshMethod === JsSIP_C.UPDATE)
        {
          this._sendUpdate();
        }
        else
        {
          this._sendReinvite();
        }
      }, expires * 500); // Half the given interval (as the RFC states).
    }
...
```
I think this fix won't break session prolongation because if there is already INVITE in progress while session timer ticks it means that we are already refreshing the session (albeit may be with different options) at about the same time. I've tested it locally in our project and it works really well.

We would really like to see this fix in JsSIP library. Thanks! :)

PR #734, opened 2021-09-15
Fix for overlapping invites while refreshing session

makstheimba:
Fix for https://github.com/versatica/JsSIP/issues/732


jmillan:
Thanks @makstheimba!

New release is out, with this fix.

makstheimba:
@jmillan Thank you very much!

PR #748, opened 2021-10-26
Generic INVITE dialog NOTIFY as basis for talk & hold NOTIFY

markusatm:
This PR is about implement support for talk and hold NOTIFY in a generic way.

Talk and hold NOTIFY are SIP NOTIFYs that are sent on a INVITE dialog. It's based on a specification originally written by Broadsoft and now maintained by Cisco.
It is a de-facto industry standard which has been implemented by several more PBX manufacturers and is based on RFC 3265.
A detailed feature explanation can be found in the specification SIPAccessSideExtensionsInterfaceSpec sections 6 and 7: https://pubhub.devnetcloud.com/media/broadsoft-docs/docs/pdf/BW-SIPAccessSideExtensionsInterfaceSpec.pdf

The change in this pull request allows the user to set an Allow-Events SIP header for a list (string array) of NOTIFY events.
This can be set:
- In the CallOptions for outgoing calls.
- In the AnswerOptions when answering incoming calls.
- Via set allow_events before a 180 Ringing gets sent on an incoming call (important!).

If a list of events is provided JsSIP will attach an Allow-Events SIP header to the initial INVITE and its responses 180 Ringing and 200 Ok (depending on incoming or outgoing call).
The Allow-Events header is only added in the initial request/response which is in accordance with RFC 3265 section 3.3.7.
If a NOTIFY request is received JsSIP will check if the client registered for events of the same type and if yes emit a 'newNotify' event and respond with 200 Ok.

I have tested this change on one of my PBXes by setting the appropriate events and verified that JsSIP correctly handles the corresponding NOTIFYs.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -2583,7 +2583,7 @@ module.exports = class RTCSession extends EventEmitter
       this.emit('newNotify', {
         originator : 'remote',
         type       : request.event.event,
-        request    : this.request
+        request    : request
       });
 
       request.reply(200);


wangduanduan:
Waiting for the feature

Issue #752, opened 2021-11-15
Enable custom From-Header for Message


gebsl:
Hello everybody,

I've got the requirement that I need to send SIP Message with different From display names, while only having one registration at the SIP proxy. In #677 such functionality was implemented for RTCCalls so I was wondering if this is possible for SIP Message.

Interface proposal would technically be the same as in #677, so just handing `fromDisplayName` via `options` object which is then forwarded as `from_display_name` via request params. As far as I've seen the handling for `from_display_name` has already been implemented in #677, so it would be a rather small change.

If that sounds feasible to you please let me know and I'll hook up a pull request.

Cheers,
gebsl

jmillan:
Hi,

Yes, feel free to prepare a PR following the one you already pointed out.

PR #753, opened 2021-11-16
Enable custom From-Header for SIP Message

gebsl:
Enables customization of From-Header for SIP Messages as suggested in #752. Implementation according to #677.

Best regards,
gebsl


jmillan:
Ready in new release, thanks!

PR #754, opened 2021-11-17
Improve matching of empty message received

BorisMorel:
Hi,

With PjSIP, empty messages received was not only "\r\n" (see below). So I change the strict matching by a regular.

Other question, why not response to this ?

Regards,
Boris.

image

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -345,6 +345,8 @@ module.exports = class Transport
     {
       logger.debug('received message with CRLF Keep Alive response');
 
+      this.socket.send('Assume that is a keep alive... So Pong');
+
       return;
     }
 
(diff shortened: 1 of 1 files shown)


jmillan:
This PR considers any message starting with CRLF a keep alive.  The original implementation is intentional, considering CRLF a keep-alive beacon.

An option would we considering one or two CRLF as keep alive: `/^(\r\n){1,2}$/`

Being double CRLF defined here https://datatracker.ietf.org/doc/html/rfc5626#section-3.5.1

UbiManu:
Hello
When do you plan to implement this fix?

jmillan:
Closed in favour of #791

PR #766, opened 2022-02-27
set iceRestart if remote sdp has more media then current on reinvite

lazedo:
* local ports for reinvite are not being setup when remote sends more media (video after audio only)


jmillan:
Is this an attempt to enable communication with a server that does not support SDP bundle? (Usage of a single transport for multiple 'm=' section)

jmillan:
Closing as by now every WebRTC server should support bundle. Re-open if needed.

PR #767, opened 2022-03-08
Fixed REFER NOTIFY parsing in accordance to RFC 3515 section 2.4.5

markusatm:
I observed failures to parse the body of a refer NOTIFY in JsSIP if the body contained more than just a status line. 
According to RFC 3515 section 2.4.5 it is allowed for additional SIP headers to be in the NOTIFY body in addition to the status line.  In accordance with that I changed the JsSIP code so the Status_Line parser is only used to parse the first line of the NOTIFY body, which according to the RFC always contains the status line. Thus the parser will no longer throw an error if there is additional data in the body after the status line.

For reference, this is the NOTIFY body that was received:
> SIP/2.0 420 Bad Extension
> Reason: SIP;cause=183;text="Call in progress"

### DIFF
--- a/lib/RTCSession/ReferSubscriber.js
+++ b/lib/RTCSession/ReferSubscriber.js
@@ -102,7 +102,7 @@ module.exports = class ReferSubscriber extends EventEmitter
       return;
     }
 
-    const status_line = Grammar.parse(request.body.trim(), 'Status_Line');
+    const status_line = Grammar.parse(request.body.trim().split('\r\n', 1)[0], 'Status_Line');
 
     if (status_line === -1)
     {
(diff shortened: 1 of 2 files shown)


ibc (approved):

ibc:
@jmillan?

jmillan:
Can you please add the corresponding test in `test/test-parser.js`?

jmillan:
BTW, thanks for the PR @markusatm and sorry for the delay. 

Once with the test we'll be able to merge it.

markusatm:
Hi @jmillan, I've added a test for status line parsing as there wasn't yet one.

Though I'm not sure if that was what you wanted as I hadn't changed anything in the parser itself. My change is to ensure that the string which is handed to the status line parser really only consists of one line.

jmillan (approved):

PR #774, opened 2022-04-21
New UA config param for global extra headers

stefang42:
Some proprietary SIP call managers require the presence of specific extra SIP headers to be present in every single request and response from a client. Currently JsSIP doesn't have a way to accomplish that, extra headers can only be added to some specific requests and responses that are directly triggered via JsSIP APIs like UA.call() or RTCSession.answer(), but many requests/responses that are triggered internally by JsSIP like ACKs or error responses don't contain these headers.

This PR adds the new UA configuration entry "extra_headers", which contains headers that should be added to every request and response, even provisional responses and ACKs. Any extra headers that are provided through the already existing APIs are added on top of these global extra headers.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -41,6 +41,9 @@ exports.settings = {
   connection_recovery_max_interval : JsSIP_C.CONNECTION_RECOVERY_MAX_INTERVAL,
   connection_recovery_min_interval : JsSIP_C.CONNECTION_RECOVERY_MIN_INTERVAL,
 
+  // Global extra headers, to be added to every request and response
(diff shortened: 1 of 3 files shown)


jmillan (approved):

ibc (approved):
Merge?

jmillan:
> Merge?

Once I prepare a new version in the docs I'll merge and add this config.

jmillan (approved):

jmillan:
Thanks @stefang42, JsSIP 3.10.0 has been released.

PR #787, opened 2022-10-28
Fix issue #706

### LINKED ISSUE #706: `_this2._request_sender.send()` causes error in `Dialog/RequestSender`
`Dialog/RequestSender` calls `_this2._request_sender.send()` in 491 status handler to make another request attempt.

But `_request_sender` doesn't exist so this line causes exception.

Looks like `this.send()` is good alternative for this line

ibc:
- Fixes #706
- `this._request_sender` doesn't exist in `Dialog/RequestSender.js`.
- Also update some deps.


jmillan (approved):

jmillan:
Ups, CHANGELOG was not updated @ibc. I'll change it and bump a new version.

PR #791, opened 2022-11-05
Transport: Handle keep alive request from server

jmillan:
Closes #754.

Be prepared to receive a keep alive request from a WS server.

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -345,8 +345,14 @@ module.exports = class Transport
     {
       logger.debug('received message with double-CRLF Keep Alive request');
 
-      // Reply with single CRLF.
-      this.socket.send('\r\n');
+      try
+      {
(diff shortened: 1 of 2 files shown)


lib/Transport.js
```diff
@@ -340,7 +340,18 @@ module.exports = class Transport
...
   _onData(data)
   {
-    // CRLF Keep Alive response from server. Ignore it.
+    // CRLF Keep Alive request from server, reply.
+    if (data === '\r\n\r\n')
+    {
+      logger.debug('received message with double-CRLF Keep Alive request');
+
+      // Reply with single CRLF.
+      this.socket.send('\r\n');
```
ibc, line 349: I'd add a catch here (not sure if try/catch or .catch()) since socket.send() can throw (or reject?).

PR #794, opened 2022-11-30
Allow to send REGISTER request without `expires`

kkozlik:
This PR adding ability to send REGISTERs without `expires` value and let the server side decide what the `expires` is.

This is done by setting `register_expires` to zero in UA config.

### DIFF
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -104,9 +104,10 @@ module.exports = class Registrator
     }
 
     const extraHeaders = this._extraHeaders.slice();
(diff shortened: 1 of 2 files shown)


lib/Registrator.js
```diff
@@ -104,10 +104,21 @@ module.exports = class Registrator
     }
 
     const extraHeaders = this._extraHeaders.slice();
+    let contactValue = `${this._contact}${this._extraContactParams}`;
```
jmillan:
```suggestion
```

PR #798, opened 2022-12-13
Sync_audioMuted and _videoMuted with local mediaStream is provided

palmtown:
This commit resolves an issue whereby when a local mediaStream is passed as an option to the answer or connect functions, the existing code does not sync the enabled status.


jmillan:
These should also emit the corresponding 'muted' events.

Issue #432, opened 2017-03-24
gatheringTimeout Missing v3.x


svarunan:
Hi 
This not a bug but a [specific feature](https://github.com/eface2face/rtcninja.js/blob/7160d2718fcf4a2ca0f90641456f35aa7294b837/dist/rtcninja.js#L540) missed out while moving to adapter.js 
We updated our JsSIP version to v3.0.0 everything works as expected but not 'gatheringTimeout'.
1. Default ICE Timeout behaviour is 10 sec. But incases when customer's Firewall blocks stun port or in slow network it causes initial outbound invite to get delayed by 10 sec. 
2. Currently i've done a work around [in ](https://github.com/versatica/JsSIP/blob/master/lib/RTCSession.js#L1565)
```js
connection.setLocalDescription(desc)
    .then(function() {
        if (connection.iceGatheringState != 'complete') {
            setTimeout(function() {
                if (connection.iceGatheringState === 'complete') {
                    return;
                }
                console.log('Timeout for ICE gathering after 2000');
                self.rtcReady = true;
                if (onSuccess) {
                    var e = {
                        originator: 'local',
                        type: type,
                        sdp: connection.localDescription.sdp
                    };
                    self.emit('sdp', e);
                    onSuccess(e.sdp);
                    onSuccess = null;
                }
            }, 2000)
        }
        console.log('iceGatheringState', connection.iceGatheringState)
    })
    .catch(function(error) {
          self.rtcReady = true;
          if (onFailure) {
              onFailure(error);
          }
          self.emit('peerconnection:setlocaldescriptionfailed', error);
    });
```
3. If will be good if `gatheringTimeout` is added back so users can pass it via `pcConfig`

ibc:
We don't want to make or maintain WebRTC workarounds. My personal goal would be to completely decouple JsSIP from the WebRTC/PeerConnection stuff, by providing an API that lets the application join both (see #427).

I will let this issue open for now. If it takes too long before we do #427, we may add a setting for this need.

sasi-plivo:
@ibc any milestone set on #427 ?

ibc:
It will be done when we get time for it within our professional priorities. Money would help prioritizing it.

tungnt1988:
anything update for this issue?
- Svarunan's solution caused the call to fail
- If not timeout, emit 'sdp' is very long
--- chrome log --
JsSIP:RTCSession new +0ms
JsSIP:RTCSession connect() +1ms
[...]

PR #800, opened 2023-01-02
Prevent repeated ICE finish

pschichtel:
This is something I noticed while implementing a rudimentary ICE gathering timeout (see #432). As you can see [here](https://github.com/VIER-CognitiveVoice/reference-webrtc-client/blob/b96d1e43c92573e8e3775aa34c5eb5a9f464192a/src/client.ts#L247-L254), I defer the call to `IceCandidateEvent.ready` by a configurable timeout. If that timeout is elapses after the iceGatheringState changed to `completed`, then `ready` will be called twice.

All other calls to the ready function in `RTCSession.js` are already conditional on the `finished` variable, I simply pushed that condition into the function.

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1959,6 +1959,10 @@ module.exports = class RTCSession extends EventEmitter
 
           const ready = () =>
           {
+            if (finished) {
+              return;
+            }
+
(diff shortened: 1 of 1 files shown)


jmillan:
thanks @pschichtel 👍

PR #805, opened 2023-02-07
Fix re-export Socket types from WebSocketInterface

Ga-hou:
When the commit id is 004d160, move the Socket and WeightedSocket of Websocket Interface.d.ts to Socket.d.ts, but JsSIP.d.ts is still importing Socket and WeightedSocket from WebsocketInterface.d.ts


jmillan:
There is no need to Re-Export a type that is already exported.

NOTE: I'm closing the issue, but please respond, or reopen if needed.

Issue #824, opened 2023-06-27
Allow set JSSIP.UA extra_headers in runtime


angeloraso:
[According to the documentation](https://jssip.net/documentation/3.10.x/api/ua/#method_set), the JSSIP.UA set method:

> Modifies the given UA configuration parameter in runtime (once started). Currently just authorization_user, password, realm, ha1, authorization_jwt and display_name can be modified.

- [ ] [Add the extra_headers property](https://github.com/versatica/JsSIP/blob/004d160968ec71982beca46d021e3cab746ac55d/lib/UA.js#L402-L442) as a valid modifiable property

cesterlizi:
Hi,  this is very useful change related  to pull request #774  that was released in version 3.1.0 since some of this headers may change during registration and without this fix we need to create a new UA every time we need to change registration headers. Can we merge this pull request and make it available for next release ?

jmillan:
This has been added to the latest release.

PR #825, opened 2023-06-27
Add extra_headers as a modificable property

angeloraso:
Issue #824


cesterlizi:
Hi, this is very useful change related to pull request https://github.com/versatica/JsSIP/pull/774 that was released in version 3.1.0 since some of this headers may change during registration and without this fix we need to create a new UA every time we need to change registration headers. Can we merge this pull request and make it available for next release ?

PR #826, opened 2023-07-05
Fix: correct SDP (new line after Duration)

Dimon70007:


ibc:
@jmillan thoughts?

jmillan:
@Dimon70007, can you show the how was the SIP message looking before your change?

jmillan:
Closing. @Dimon70007, feel free to reopen if you point to the corresponding spec justifying this change.

PR #828, opened 2023-07-17
Update UPDATE method

OliverGarfield:
The update is usually done before the invite request is answered, so the processing of 1XX states is added

### DIFF
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -1578,7 +1578,8 @@ module.exports = class RTCSession extends EventEmitter
           }
           break;
         case JsSIP_C.UPDATE:
-          if (this._status === C.STATUS_CONFIRMED)
+          if (this._status === C.STATUS_1XX_RECEIVED ||
+              this._status === C.STATUS_CONFIRMED)
           {
             this._receiveUpdate(request);
           }


OliverGarfield:
image

OliverGarfield:
@jmillan Hi?

ibc:
It's summer here and hence vacation days. We'll come to this and many other pending work on next weeks.

OliverGarfield:
> It's summer here and hence vacation days. We'll come to this and many other pending work on next weeks.

Ok,I'm sorry to disturb your vacation.

stefang42:
Hi @OliverGarfield! Sorry, but I think your modification is not quite correct. The UAS ("Beale" in your diagram) is not in state STATUS_1XX_RECEIVED when it receives the UPDATE request, because it's the other way around: the UAC ("Vigenere") that sent the INVITE received the provisional response. After receiving the INVITE and sending the provisional responses the UAS will be in state STATUS_WAITING_FOR_ANSWER, not STATUS_1XX_RECEIVED, when it receives the UPDATE.

(Also, looking at https://www.rfc-editor.org/rfc/rfc3311.html#section-5.2, I'm not sure JsSIP is already handling UPDATE requests correctly in every case, depending on the offers/answers already exchanged or not...)

OliverGarfield:
@stefang42 You're welcome! Thank you for your response and for providing important clarifications regarding SIP (Session Initiation Protocol) and RFC 3311. I have read your points and understand your explanation.

You are absolutely correct in pointing out that the UAS ("Beale" in your diagram) typically enters the "STATUS_WAITING_FOR_ANSWER" state after receiving an INVITE request and sending provisional responses. In RFC 3311, it is indeed stated that the "UPDATE" request is replied to after entering the "STATUS_WAITING_FOR_ANSWER" state, contingent on the presence of a "PRACK" message. However, it is a valid consideration whether one should reply to the "UPDATE" in the "STATUS_1XX_RECEIVED" state if no "PRACK" message is present.

This is indeed a valid point, and the decision to respond to an "UPDATE" in the "STATUS_1XX_RECEIVED" state in the absence of a "PRACK" message may depend on specific implementations and requirements. It's important to consider the particular use case and how different implementations handle this scenario.

stefang42:
Hi @OliverGarfield! Sorry, maybe I did not express myself clearly before. I'm not necessarily saying your current patch is wrong. While I have never seen this scenario personally, there may indeed be cases where the caller that sent the INVITE may receive an UPDATE request after receiving a provisional response (i.e. the caller's JsSIP RTCSession is in state STATUS_1XX_RECEIVED), depending on whether offers/answers have been exchanged (though PRACK is currently not supported by JsSIP I think). 

I'm just saying that in my opinion the current patch is not sufficient, because it doesn't cover the situation where the callee receives an UPDATE before generating the final response to the INVITE it received, i.e. the situation outlined in the diagram you attached. There on the callee side JsSIP's RTCSession will be in state STATUS_WAITING_FOR_ANSWER when it receives the UPDATE, which is a valid scenario, e.g. when no offer has been exchanged yet. So I think your patch should also cover that state, not only STATUS_1XX_RECEIVED.

jmillan:
Agreed with @stefang42, this PR needs more coverage for the given scenarios, at least.

jmillan:
Hi,

We don't support PRACK. I'm closing this PR since it's not complete as per the comments. Feel free to open a new one with the exact case (please include full logs) where JsSIP may be failing handling the UPDATE.

PR #833, opened 2023-09-14
feat: fire sipEvent on RTCSession when STATUS_WAITING_FOR_ANSWER

wangduanduan:
when jssip receivce NOTIFY (in dialog) before call answer,  jssip reply with 403 Wrong Status.

but we want to use in-dialog NOTIFY

the code below is my in app , no in jssip

```js
myPhone.on('newRTCSession', ({ session, originator, request })=>{

    session.on('sipEvent', (e) => {
       
    })
})
```

### DIFF
--- a/lib/RTCSession.d.ts
+++ b/lib/RTCSession.d.ts
@@ -176,6 +176,11 @@ export interface IncomingAckEvent {
   ack: IncomingRequest;
 }
 
+export interface SipEvent {
+  event: string;
+  request: IncomingRequest;
+}
+
 // listener
 export type PeerConnectionListener = (event: PeerConnectionEvent) => void;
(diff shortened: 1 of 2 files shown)


markusatm:
There's already a pending pull request I created for this two years ago: #748 
I recommend using that one since it fully implements it and makes it optional. Always simply accepting a call if a "talk" notify arrives is a security risk.

wangduanduan:
@markusatm but why they pending your pull request for two years!  😂.  #748

ibc:
We will address pending PRs when we are less busy, but indeed this PR is no valid due to above given rationale.

wangduanduan:
@markusatm 

I just need the sipEvent in dialog, maybe not answer call

Issue #799, opened 2023-01-02
Getting error on npm start in angular as `Module '"events"' has no exported member 'Listener'.`


dnyaneshwerEquentis:
I am getting following error in my angular project.

Error: node_modules/jssip/lib/RTCSession.d.ts:1:23 - error TS2305: Module '"events"' has no exported member 'Listener'.

I am using angular 14 and typescript version 4.8.4

Any help on this would be much appreciated?

tzahari:
I found a workaround form https://stackoverflow.com/a/74989759
But this is only a workaround and not a fix.

jmillan:
This has been fixed in the latest release.

PR #837, opened 2023-10-26
Fix type definitions to not import `Listener` from `events` (fixes: #799)

jlaine:
The type definitions provided by `@types/events` clash with those by `@types/node`, resulting in the error:

         Module '"events"' has no exported member 'Listener'.

In practice this makes it impossible to use JsSIP in any Angular project, resulting in issues such as #799.

Looking at the places where `Listener` is being imported, we are not gaining anything by using this type definition which is extremely generic. We can do better by using a more specific type definition.


jmillan:
Makes sense, thanks @jlaine :+1:

Issue #821, opened 2023-06-01
Change type of request in OutgoingRTCSessionEvent to OutgoingRequest


SergioFatalist:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `jssip@3.10.0` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/jssip/lib/UA.d.ts b/node_modules/jssip/lib/UA.d.ts
index 13798d3..1e51f3d 100644
--- a/node_modules/jssip/lib/UA.d.ts
+++ b/node_modules/jssip/lib/UA.d.ts
@@ -55,7 +55,7 @@ export interface IncomingRTCSessionEvent {
 export interface OutgoingRTCSessionEvent {
   originator: Originator.LOCAL;
   session: RTCSession;
-  request: IncomingRequest;
+  request: OutgoingRequest;
 }
 
 export type RTCSessionEvent = IncomingRTCSessionEvent | OutgoingRTCSessionEvent;
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

AlijiEmmanuel-dev:
I was just about to report this issue. It has not been fixed.

ibc:
Fixed in https://github.com/versatica/JsSIP/pull/846

PR #846, opened 2024-02-07
Fixed some broken TypeScript definitions in RTCSession.d.ts and UA.d.ts

stefang42:
This commit fixes some type definitions that unfortunately were broken by some of the recent merges of some pull request.

Commit d547d74be9f757dc23a1065670c8d10515152f17 (Use built-in MediaStreamConstraints type (#809)) removed the interface MediaConstraints, but failed to adjust the definitions of RTCSession's MuteListener and the mute, unmute and isMuted functions.
This leads to errors when transpiling with TypeScript:
TS2552: Cannot find name MediaConstraints. Did you mean MediaTrackConstraints?

Commit 901574f75e06645776a4fad5793903d3c8488844 (Allow to configure from-tag in register (#793)) introduced UAConfiguration.register_from_tag_trail with an invalid type that causes an error with TypeScript:
TS8020: JSDoc types can only be used inside documentation comments.

I also took the opportunity to fix a wrong type definition in OutgoingRTCSessionEvent, for which there is also an open issue #821.

### DIFF
--- a/lib/RTCSession.d.ts
+++ b/lib/RTCSession.d.ts
@@ -171,6 +171,11 @@ export interface IncomingAckEvent {
   ack: IncomingRequest;
 }
 
+export interface MediaStreamTypes {
+  audio?: boolean;
+  video?: boolean;
+}
+
 // listener
 export type GenericErrorListener = (error: any) => void;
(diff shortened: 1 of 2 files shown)


ibc (approved):

ibc:
Thanks

Issue #861, opened 2024-06-21
On iOS, navigator.mediaDevices is `undefined` unless page is served in a secure context, but Errors are swallowed


pjkundert:
OK, so part one of the issue was this:

iOS will refuse to allow Javascript code to access navigator.  In RTCSession.js, when we call

`return navigator.mediaDevices.getUserMedia(mediaConstraints)`

the navigator.mediaDevices is `undefined` -- unless you're accessing the code in a "secure context":  via https://... or file:///

This is not checked, causing an exception -- except since the exception is not thrown by the call itself, but by the code, it trickles down to the final `.catch((error) =>`, which neither logs nor re-throws the Error, it just swallows it.

I'll submit an Issue and a Pull for this, of course, but there are a couple of other things, too...

_Originally posted by @pjkundert in https://github.com/versatica/JsSIP/discussions/860#discussioncomment-9843435_

PR #862, opened 2024-06-21
Initial pass at fixing navigator.mediaDevices

pjkundert:
First pass; chatty and probably wrong...

fixes: #861


pjkundert:
I have a demo that can serve up a simple JsSIP interface via a crude https session w/ self-signed certificates:

https://github.com/pjkundert/jsSip-demo

jmillan:
This is already fixed in new release.

We are terminating the session and emitting `failed` event now if the method cannot complete successfully.

Issue #556, opened 2018-12-23
blind call transfer improvement


ikq:
I found problem with the blind call transfer scenario:

1. 1st phone send REFER to 2nd
2. 2nd: REFER accepted, and start calling to 3rd
3. According RFC 5589 1st phone should not here send BYE to 2nd until receive NOTIFY with final call result.
   But ... 1st phone send BYE to 2nd !
4. Now then 3rd phone accept incoming from 2nd phone call, in the 2nd will try send NOTIFY to 1st.
   But it will failed, then it will close 2nd-3rd phone call, and there will be exception.....
   Seems the difficult case did not implemented.

According description CISCO: "https://www.cisco.com/c/en/us/td/docs/ios/voice/sip/configuration/guide/12_4t/sip_12_4t_book/sip_cg-call_x-fer.pdf", page 8. Figure 2.
It's still possible send REFER and then immediately BYE
Better in the case then 2nd phone continue to call to 3rd, and then establish the call without problem.

Proposed fix in file: ReferNotifier.js
Don't send NOTIFY if dialog (that receive REFER) is terminated.

notify(code, reason)
  {
    debug('notify()');

    if (this._active === false)
    {
      return;
    }
	
	if( this._session.status === this._session.C.STATUS_TERMINATED) //     Proposed fix: 
	{                                                                                                      //      Don't try send NOTIFY
		this._active = false;                                                                 //      if dialog is terminated.
		return;                                                                                    //
	}                                                                                                      //

ikq:
Yes, after 1st phone send BYE, it don’t receive NOTIFY.

Problem with 2nd phone, seems there some exception when notifier try to send NOTIFY.
I can provide logs from 2nd phone, if you want.

jmillan:
Ok, then provide us the full JsSIP logs for the 2nd phone.

ikq:
Hi !

Here the logs of my test phone and jssip

phone1 (user web1), and phone2 (user web2), call transferred to phone3 (user web3)

I send also build jssip.js, because I think added fix of ICE timeout can a little change lines numbers.

Scenario


  1.  Phone1  (user web1) call to phone2 user web2,
  2.  Phone2 answer, call is established.
  3.  Phone1 set call on hold.
  4.  Phone1 send to Phone2 REFER with request to call phone3 (user web3)
  5.  REFER is accepted, Phone2 starts calling to Phone3
  6.  (!) Phone1 hangup call (send BYE) to phone2
[...]

PR #878, opened 2024-10-22
ReferNotifier sendRequest ignored #556

miguelnog19itcenter:
Scenario to be solved:
A is on communication with B
B transfers the call to C (blind transfer)
B hangups the call with A when trying is received
C answers the call and the exception is triggered, killing the call

The exception is triggered whenever A is trying to notify B about the refer status.


jmillan:
May you post here what exception was being triggered? Can you provide full logs of JsSIP?

jmillan:
Also please add the full JsSIP logs.

Issue #782, opened 2022-07-27
Vue3 + jsSIP cannot be used properly


vssir:
Hello everyone, I found a problem, the project is based on VUE3 + TS + Vite, the following error occurs when SIP users register after compiling.

Is it a BUG?

console err:
![image](https://user-images.githubusercontent.com/10388068/181239697-c9a9b79f-8781-4630-aa46-17848a098f58.png)

position:
![image](https://user-images.githubusercontent.com/10388068/181239802-76629a35-3f75-42ff-92ca-ba5cb57b5d7b.png)

jssip source code:
![image](https://user-images.githubusercontent.com/10388068/181240249-02a83610-9d63-4c62-9d23-b5e307100aa9.png)

boren4321:
hi, how to resole it?pls

antant7:
The issue persists - Quasar (Vue3) + TS + Vite - only after project build. No errors in dev mode.

Temporary workaround - modify line **\node_modules\jssip\lib-es5\NameAddrHeader.js:37** 
from ```if (!uri || !(uri instanceof URI))``` to ```if (!uri || (URI && !(uri instanceof URI)))```.

After updating the package, you'll need to re-edit it again. It's very inconvenient, but otherwise SIP calls stop working.

shevtsov-max:
My [pull request](https://github.com/versatica/JsSIP/pull/889) for a change was rejected, so I had to create a "forker" in my profile in order to use the library.

edit package.json
`"jssip": "https://github.com/shevtsov-max/JsSIP/tarball/master",`

It helped me.

PR #889, opened 2024-12-27
Update NameAddrHeader.js

shevtsov-max:
https://github.com/versatica/JsSIP/issues/782

### DIFF
--- a/lib/NameAddrHeader.js
+++ b/lib/NameAddrHeader.js
@@ -24,7 +24,7 @@ module.exports = class NameAddrHeader
   constructor(uri, display_name, parameters)
   {
     // Checks.
-    if (!uri || !(uri instanceof URI))
+    if (!uri || (URI && !(uri instanceof URI)))
     {
       throw new TypeError('missing or invalid "uri" parameter');
     }


lib/NameAddrHeader.js
```diff
@@ -24,7 +24,7 @@ module.exports = class NameAddrHeader
   constructor(uri, display_name, parameters)
   {
     // Checks.
-    if (!uri || !(uri instanceof URI))
+    if (!uri || (URI && !(uri instanceof URI)))
```
ibc, line 27: This is not gonna fix anything. If `URI` doesn't exist in window/global then `if (URI)` will throw anyway. 

This is not a valid fix. 

The problem is that `URI.js` loads `Utils.js` and `Utils.js` also loads `URI.js` so at the time `URI.js` is loaded it's empty. This is a problem we had in the past and requires a mayor refactor. This PR is not fixing it at all so I'm closing it.

shevtsov-max: In order to use the library, I need this fix. I did it through my profile if you can't do this check. I'm not the only one with the problem, and it's been a long time, judging by the open issue.

ibc: I've been very specific in my comment above and you are not saying anything about it.

shevtsov-max: You've given a good answer, but it doesn't solve the problem right now. You wrote "This is a problem we had in the past and requires a mayor refactor", but when will it be available? I'm offering a temporary solution that will help now, and at least many people will be able to run the library. And then you can solve the problem as you described.

ibc: What I meant is that your workaround cannot properly work despite it seems that it works. The problem is that, using Vue require() implementation there is a deadlock situation and one of the calls to require("./URI.js") is literally importing nothing and that's why you get undefined URI.

PR #895, opened 2025-01-30
React-native-support

cyrenity:


lib/RTCSession.js
```diff
@@ -17,7 +17,9 @@ const RTCSession_ReferNotifier = require('./RTCSession/ReferNotifier');
 const RTCSession_ReferSubscriber = require('./RTCSession/ReferSubscriber');
 const URI = require('./URI');
 
-const logger = new Logger('RTCSession');
+const { registerGlobals } = require('react-native-webrtc');
```
hthetiot, line 20: You introduced Dependencies without an package and I don't think this project should depend on react-native-webrtc that does not make sense.

lib/RTCSession.js
```diff
@@ -833,7 +835,7 @@ module.exports = class RTCSession extends EventEmitter
 
         if (status_code && (status_code < 200 || status_code >= 700))
         {
-          throw new TypeError(`Invalid status_code: ${status_code}`);
+          return new TypeError(`Invalid status_code: ${status_code}`);
```
ibc, line 838: What?? Sorry, this doesn't make any sense.

lib/RTCSession.js
ibc: Correct.

ibc:
Sorry, this is not the way to go.

lib/RTCSession.js
hthetiot: That new, usually those script kiddies use try catch nothing because they want to ignore those pesky errors the return new Error is a new one for me ahah.

ibc: Yes, let's assume errors don't exist and write optimistic code. It's fantastic.

PR #905, opened 2025-03-12
Custom Security-Websocket-Protocol Support

khosbilegt:
It is possible to modify the Security-Websocket-Protocol header for Websockets and JsSIP has no support for it at the moment (hardcoded as ['sip']). 

This change adds an optional `_appendedProtocols` parameter for WebSocketInterface that appends the given array of strings to the Security-Websocket-Protocol header.


ibc:
Sorry, the standard `Sec-WebSocket-Protocol` in RFC 7118 _"The WebSocket Protocol as a Transport for SIP"_ is literally "sip":

https://datatracker.ietf.org/doc/html/rfc7118#section-4.1

JsSIP implements it and hence it must use `Sec-WebSocket-Protocol: sip` rather than something else.

Also, your PR is changing the syntax of the code, which of course it shouldn't do.

PR #925, opened 2025-08-26
Fix JsSIP 3.7.1 doesn't allow sending DTMF during EarlyMedia #689

giacco:
The patch introduces two changes:

### Early dialog handling

* When no confirmed dialog (this._dialog) is present, the code now attempts to send the request through the first available early dialog.

* Added safety checks to ensure _earlyDialogs is not empty and that the selected dialog supports sendRequest.

### CSeq update in Dialog.update()

* The patch updates the dialog’s CSeq based on the received SIP message, preventing desynchronization and improving compliance with SIP dialog state handling.

### DIFF
--- a/lib/Dialog.js
+++ b/lib/Dialog.js
@@ -83,7 +83,7 @@ module.exports = class Dialog
       this._remote_uri = message.parseHeader('to').uri;
       this._remote_target = contact.uri;
       this._route_set = message.getHeaders('record-route').reverse();
-      this._ack_seqnum = null;
+      this._ack_seqnum = this._local_seqnum;
 
     }
 
@@ -202,8 +202,9 @@ module.exports = class Dialog
 
     if (!this._local_seqnum) { this._local_seqnum = Math.floor(Math.random() * 10000); }
 
+    // CANCEL and ACK must use the same sequence number as the INVITE.
     const cseq = (method === JsSIP_C.CANCEL || method === JsSIP_C.ACK) ?
-      this._local_seqnum :
+      this._ack_seqnum :
       this._local_seqnum += 1;
 
     const request = new SIPMessage.OutgoingRequest(
--- a/lib/RTCSession.js
+++ b/lib/RTCSession.js
@@ -2076,10 +2076,6 @@ module.exports = class RTCSession extends EventEmitter
       {
         early_dialog.update(message, type);
         this._dialog = early_dialog;
-        // When an early dialog is promoted to a confirmed dialog (e.g., after calling an IVR menu),
-        // the local CSeq must be synchronized
-        // with the value from the 200 OK response to keep the dialog sequence in sync.
-        this._dialog._local_seqnum = message.cseq;
         delete this._earlyDialogs[id];
 
         return true;


dist/jssip.js
```diff
@@ -598,6 +598,14 @@ module.exports = /*#__PURE__*/function () {
         // RFC 3261 13.2.2.4.
         this._route_set = message.getHeaders('record-route').reverse();
       }
+      var cseq = message.cseq ? parseInt(message.cseq, 10) : null;
```
jmillan, line 601: `message.cseq` is guaranteed to be there. `sanityCheck.js` use from `UA.js` ensures it.

lib/Dialog.js
```diff
@@ -142,6 +142,20 @@ module.exports = class Dialog
       // RFC 3261 13.2.2.4.
       this._route_set = message.getHeaders('record-route').reverse();
     }
+
+    const cseq = message.cseq ? parseInt(message.cseq, 10) : null;
```
jmillan: `message.cseq` is guaranteed to be present. `sanityCheck.js` used from `UA.js` ensures it.

lib/RTCSession.js
```diff
@@ -1420,6 +1420,20 @@ module.exports = class RTCSession extends EventEmitter
   {
     logger.debug('sendRequest()');
 
+    if (!this._dialog)
```
jmillan: Let's reorder this and try first to send it over `this._dialog` and if not present use an early dialog if present.

jmillan: Since our main intention here is to use the stablished dialog.

lib/Dialog.js
jmillan: Indeed this should not be needed at all as the cseq must be equal to the original request. Did you see any problem with the current approach?

giacco: The reason I introduced the update of `_local_seqnum` / `_remote_seqnum` from `message.cseq` is that during testing with an IVR sending `183 Session Progress` with early media, calls would consistently drop after ~11–12 seconds if this assignment was not done. 

Even though the DTMF INFO requests were being delivered, the dialog state was not stable and the PBX terminated the session. Updating the dialog’s seqnum here prevented those drops and kept the call alive.

lib/RTCSession.js
giacco: Totally agree

lib/Dialog.js
jmillan: It's worth it investigating this because the responses of a request have the same seq num as the request itself, and the dialog's seq num is already set with the request one so.. this lines are fixing something that should be fixed somewhere else. We should get to the root of the error.

jmillan: Here are local and remote sequence numbers set upon dialog construction:

https://github.com/versatica/JsSIP/blob/master/lib/Dialog.js#L61
https://github.com/versatica/JsSIP/blob/master/lib/Dialog.js#L81

giacco: I have moved the local CSeq update to RTCSession.js, where the early dialog is promoted to dialog.
Without it, calls with IVR and early media drops when the remote party answered aftert 12 / 13 sec max.

@jmillan   please let me know if you think there’s a more appropriate place to handle this CSeq alignment, but based on testing this adjustment keeps the dialog stable without affecting the standard dialog update flow.

jmillan: Since the change in `Dialog.js` should not be needed, yet your scenario seems to fail if not set. Could you please share the JsSIP logs (including SIP traces) with and without your change so we do know why is it failing in such case?

giacco: I’m attaching both the JsSIP console logs and the WebSocket SIP message captures (exported via the browser HAR file ,  specifically the _webSocketMessages[] array ).


* File 1 contains the capture **with** the CSeq update. [1._Flow_with_Cseq_update.txt](https://github.com/user-attachments/files/24105755/1._Flow_with_Cseq_update.txt)

* File 2 contains the capture **without** the CSeq update. [2._Flow_with_NO_Cseq_update_(Broken).txt](https://github.com/user-attachments/files/24105757/2._Flow_with_NO_Cseq_update_.Broken.txt)

The call flow is identical in both files:

1. SIP account ibJfGT201 (running the WebApp using JsSIP) calls the IVR (777).

2. The IVR replies with 183 Session Progress with early media.

3.  At this point, ibJfGT201 sends DTMF “1” to the IVR.

4. The IVR forwards the call to 202.

5. Extension 202 answers.

6. Communication between ibJfGT201 and 202 starts.

 Issue:
If the CSeq is not updated, the call stays up for about 12–13 seconds and then drops with “No user responding”.


Asterisk logs from the PBX

On the Asterisk side I see
```
WARNING[15159] chan_sip.c: Retransmission timeout reached on transmission mbsg1f0dln4kc23o4i49 for seqno 4045 (Critical Response)
Packet timed out after 12804ms with no response
WARNING[15159] chan_sip.c: Hanging up call mbsg1f0dln4kc23o4i49 - no reply to our critical packet

```

Followed by the channels leaving the bridge and the call being terminated.

Let me know if you need additional captures

jmillan: Thanks @giacco, I see the problem now:

|Action|CSeq in this PR|CSeq this master|
|-|-|-|
|-> INVITE | 6356|4044|
|<- 401|6356|4044|
|-> INVITE | 6357|4045|
|-> 183 | 6357|4045|
|-> NOTIFY | 6358|4046|
|<- 200 | 6357|4045|
|-> ACK | 6357|4046|

Here the problem in master: The `NOTIFY` increases the dialog cseq so later on the `ACK` is sent with a value of `4046` rather than the `4045` of the authenticated `INVITE`.

The approach in this PR works, but now imagine you are sending another `NOTIFY` after the `ACK`. Since the internal cseq was reset to 6357, a `NOTIFY` with a `CSeq` 6358 would be created (just as the one created earlier), and this could generate further problems in different scenarios.

Can you please try this patch which I believe is the correct way to handle it? Basically we are assigning an `_ack_seqnum` out of the 1XX or 2XX response creating the dialog, and then use that `CSeq` for generating the `ACK` or the `CANCEL`.

```diff
diff --git a/lib/Dialog.js b/lib/Dialog.js
index 18d0dd7..baa5ed6 100644
--- a/lib/Dialog.js
+++ b/lib/Dialog.js
@@ -83,7 +83,7 @@ module.exports = class Dialog
       this._remote_uri = message.parseHeader('to').uri;
       this._remote_target = contact.uri;
       this._route_set = message.getHeaders('record-route').reverse();
-      this._ack_seqnum = null;
+      this._ack_seqnum = this._local_seqnum;

     }

@@ -202,8 +202,9 @@ module.exports = class Dialog

     if (!this._local_seqnum) { this._local_seqnum = Math.floor(Math.random() * 10000); }

+    // CANCEL and ACK must use the same sequence number as the INVITE.
     const cseq = (method === JsSIP_C.CANCEL || method === JsSIP_C.ACK) ?
-      this._local_seqnum :
+      this._ack_seqnum :
       this._local_seqnum += 1;

     const request = new SIPMessage.OutgoingRequest(
```

giacco: I have tried your patch and it works !.
Here the log, with the same flow as before, but with cseq managed as you suggested.
[3.SameFlowWithProperCseqHandle](https://github.com/user-attachments/files/24169531/3.Flow_with_your_fix.txt)

jmillan: Thanks @giacco,

Would you please update this PR to not include any file in /dist folder, just those under /lib?

Thanks!

giacco: I’ve updated the PR and removed all build artifacts.
The PR now includes only source files under /lib

jmillan:
Change applied in the new release. Thanks a lot @giacco :-)

PR #929, opened 2025-09-12
allow reg id and reg call id to be passed

rakesh-voiger:
Currently, upon page refresh, the registration request being sent does not use the same Call-Id header in the REGISTER request, hence the backend (FreeSWITCH) assumes it's a different connection and keeps the old connection also open for a while (when multi connection per user is enabled). The old connection goes off after a while, but for some time it will be in memory and calls will be attempted to these connections.

Hence this pr, which will allow the client to explicitly pass the call-id and reg id for the registration requests, so frontend can use the same reference even after the reloads.

### DIFF
--- a/lib/Registrator.js
+++ b/lib/Registrator.js
@@ -13,7 +13,7 @@ module.exports = class Registrator
   constructor(ua, transport)
   {
     // Force reg_id to 1.
-    this._reg_id = 1;
+    this._reg_id = ua.configuration.registration_id || 1;
 
     this._ua = ua;
     this._transport = transport;
@@ -22,7 +22,7 @@ module.exports = class Registrator
     this._expires = ua.configuration.register_expires;
 
     // Call-ID and CSeq values RFC3261 10.2.
-    this._call_id = Utils.createRandomToken(22);
+    this._call_id = ua.configuration.registration_call_id || Utils.createRandomToken(22);
     this._cseq = 0;
 
     this._to_uri = ua.configuration.uri;


rakesh-voiger:
@ibc Your thoughts ?

jmillan:
> @ibc Your thoughts ?

Please avoid direct mentions.

You can unregister the UA on page unload using the window event [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)

rakesh-voiger:
> > @ibc Your thoughts ?
> 
> Please avoid direct mentions.
> 
> You can unregister the UA on page unload using the window event [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)

Apologies for the direct mention.

unregister is not an option for me, as i want to give user a chance to comeback within few seconds without unregistering and registering back. the server has timeout after which the registration expires.

ibc:
> unregister is not an option for me, as i want to give user a chance to comeback within few seconds without unregistering and registering back

Unregister and then register again is the correct approach here despite you say it's not an option for you. And as Jose said, you can unregister the UA on page unload using the window event [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event).

In addition, FreeSwitch should be able to delete the registration of a UA when its WebSocket connection is closed, but that's another story.

Sorry, we cannot make these low level IDs configurable, that's not the right approach and that's something that doesn't exist in any device in SIP world.

rakesh-voiger:
> > unregister is not an option for me, as i want to give user a chance to comeback within few seconds without unregistering and registering back
> 
> Unregister and then register again is the correct approach here despite you say it's not an option for you. And as Jose said, you can unregister the UA on page unload using the window event [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event).
> 
> In addition, FreeSwitch should be able to delete the registration of a UA when its WebSocket connection is closed, but that's another story.
> 
> Sorry, we cannot make these low level IDs configurable, that's not the right approach and that's something that doesn't exist in any device in SIP world.

got it, thanks for the information

PR #930, opened 2025-09-24
Support for WebRTC overrides 3.10.1

Samarthya2912:


jmillan:
The PR has no description at all.

JsSIP uses:

- window.RTCPeerConnection
- navigator.mediaDevices.getUserMedia

If you want to rewrite those there is no need to do it inside JsSIP.

PR #942, opened 2025-12-31
add optional sip call-id

eschmidbauer:


ibc:
Call-Id value is something internal at SIP protocol level. There is no reason for letting the user application choose it.

eschmidbauer:
there is benefit from letting the user customize it. For example setting a matching call-id in REGISTER to connect a suspended INVITE with same call-id. 
This change is backward compatible and RFC compliant.

ibc:
> For example setting a matching call-id in REGISTER to connect a suspended INVITE with same call-id.

I don't know what that is and I've never seen anything about that in any SIP RFC specification.

eschmidbauer:
https://datatracker.ietf.org/doc/html/rfc3261#section-8.1.1.4
Use of cryptographically random identifiers ([RFC 1750](https://datatracker.ietf.org/doc/html/rfc1750) [[12](https://datatracker.ietf.org/doc/html/rfc3261#ref-12)]) in the
   generation of Call-IDs is RECOMMENDED.  Implementations MAY use the
   form "localid@host".  Call-IDs are case-sensitive and are simply
   compared byte-by-byte.
   
   
it says "...generation of Call-IDs is RECOMMENDED." therefore it is not illegal to have the call-id customized

eschmidbauer:
it is also beneficial for users to set the call-id for logging/tracing purposes.

jmillan:
> For example setting a matching call-id in REGISTER to connect a suspended INVITE with same call-id.

What is this about? I don't recall it from any spec.

> it is also beneficial for users to set the call-id for logging/tracing purposes.

There is already a `from_tag` customization for that exact purposes. Currently implemented for registration. More [here](https://github.com/versatica/JsSIP/pull/793/changes).

eschmidbauer:
> > For example setting a matching call-id in REGISTER to connect a suspended INVITE with same call-id.
> 
> What is this about? I don't recall it from any spec.

This is not from a spec but a common practice in mobile setups. You can suspend a SIP INVITE on server, send push notification (with the sip call-id of invite), then when the SIP REGISTER arrives with matching sip call-id, the INVITE is resumed and sent to the UAC. The spec i mentioned was related to setting the sip call-id in registers, it is RFC complaint to customize the sip call-id, also the code in the PR is backward compatible. I think folks would benefit from this change but we are fine running our own fork which implements this, please close the PR if you do not want it merged. 

> 
> > it is also beneficial for users to set the call-id for logging/tracing purposes.
> 
> There is already a `from_tag` customization for that exact purposes. Currently implemented for registration. More [here](https://github.com/versatica/JsSIP/pull/793/changes).

ibc:
> it says "...generation of Call-IDs is RECOMMENDED." therefore it is not illegal to have the call-id customized

This is a note for the SIP device implementors. It says nothing about the user being able to "customize" the call-id.

eschmidbauer:
> > it says "...generation of Call-IDs is RECOMMENDED." therefore it is not illegal to have the call-id customized
> 
> This is a note for the SIP device implementors. It says nothing about the user being able to "customize" the call-id.

It is implied because the RFC only recommends auto-generation. I will close PR

ibc:
No need to close it, we are just discussing. I am on vacation, let's us come back to this soon.

PR #949, opened 2026-01-13
RtcSession: Add missing typedef import for IncomingRequest

orgads:


jmillan:
Shouldn't this be part of RTCSession.d.ts anyway?

I'm planing to move JsSIP to Typescript (I've already started the process), so there will be no need for this ones.

orgads:
Ok

PR #950, opened 2026-01-14
Added esbuild task for building minified version

Hecatron:
Hi,
I was looking to see if there was a minified version of jssip that could be downloaded
but it looks as if the scripts for generating one were removed from the source.

I've added one to npm-scripts.js / package.json that runs esbuild
This is quite a bit faster than the original Browserify
I've also added /dist/ to .gitignore since normally you wouldn't want to include minified code into the github repo directly

What should be possible after this, is to add a line to the release function to run the build
then you could add an entry to package.json to also copy /dist/jssip.min.js into the final package published to npmjs
This is the way I think most other packages tend to do it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /node_modules/
+/dist/
--- a/npm-scripts.js
+++ b/npm-scripts.js
@@ -3,50 +3,59 @@ const path = require('path');
 const process = require('process');
 const { execSync } = require('child_process');
 const { version } = require('./package.json');
+const pkg = require('./package.json');
+const { resolve } = require("path");
+const { build } = require("esbuild");
+const { readFile, writeFile } = require("fs/promises");
 
 const task = process.argv.slice(2).join(' ');
 
 const ESLINT_PATHS = [ 'src', 'test' ].join(' ');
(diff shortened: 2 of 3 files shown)


ibc:
There isn't any good reason in 2026 to generate minified JS libraries. Those who use JsSIP are supposed to integrate it into a larger application built with React, Vue, webpack, etc, rather than integrating JsSIP into their website as if it was year 2005.

Hecatron:
There are situations where a minified file is useful, typically in cases outside of js frameworks
Or if you want to download via a cdn

  * Use of a non js framwework such as blazor - [JSImport]
    https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/?view=aspnetcore-10.0
  * Use in rust / web assembly
    https://book.leptos.dev/web_sys.html

jmillan:
@Hecatron we recently removed the bundle generation, but indeed you have a point. There always was, and there is no cost on having it. Let me take over the PR in order to consolidate it a bit.

jmillan:
Closed in favour of #951

Hecatron:
Thanks for the above, I'll close this pull request

PR #955, opened 2026-01-22
Modernize

jmillan:


jmillan:
> Have you tried "npm pack" to generate a "npm tgz" library and then install it somewhere else via "npm install tgz" and see its content in node_modules/jssip?

I've reduce it to the minimum needed, as we do in the rest of libraries.

Issue #970, opened 2026-04-12
Refer-To URI headers copied to target URI of outgoing INVITE (since fix for #962)


karstenluedtke:
# Bug Report
## Your environment

- JsSIP version: 3.13.6

## Issue description
The fix for #962 removed the toAor() call from `session.connect(request.refer_to.uri, options, initCallback);`.
As intended, the URI parameters of the refer_to URI are now included in the outgoing INVITE,
but now the refer_to URI headers are copied, too.

In my understanding, this is wrong.
The URI headers in the refer_to URI shall be used to construct the INVITE (as it is done for the Replaces header
a few lines above), but they shall not be forwarded to the target in the request-URI or To header.
We don't want the target to include those headers into the messages originating from the target.

I would suggest following change:
```
diff --git a/src/RTCSession.js b/src/RTCSession.js
index 47f8848..63fff69 100644
--- a/src/RTCSession.js
+++ b/src/RTCSession.js
@@ -2314,7 +2314,9 @@ module.exports = class RTCSession extends EventEmitter {
                                options.extraHeaders.push(`Replaces: ${replaces}`);
                        }
 
-                       session.connect(request.refer_to.uri, options, initCallback);
+                       let target_uri = request.refer_to.uri.clone();
+                       target_uri.clearHeaders();
+                       session.connect(target_uri, options, initCallback);
                }
 
                function reject() {
```

jmillan:
Thanks for reporting.  Can you please provide with a PR?

PR #971, opened 2026-04-24
removing refer_to uri headers when creating target uri for INVITE - #970

karstenluedtke:
Pull request for my change concerning issue #970:

It will prevent that the URI headers in the refer_to URI are forwarded to the target in the request-URI or To header.


jmillan (approved):
Thanks, I'll release a new version early next week.

jmillan:
@karstenluedtke, could you please fix the lint issue.

Run `npm run lint` locally.

jmillan:
Available in latest release, thanks @karstenluedtke 👌

